PR #526, opened 2019-01-10
sqlite: allow to use drop_constraint, drop_column and create_constraint

vvvrrooomm:
These changes implement drop_constraint, drop_column and create_constraint on sqlite. 

It might require some advice on rework (that I'm happy to do) since it feels that abstraction layers were broken.

### DIFF
--- a/alembic/ddl/sqlite.py
+++ b/alembic/ddl/sqlite.py
@@ -2,7 +2,7 @@ import re
 
 from .impl import DefaultImpl
 from .. import util
-from sqlalchemy import Table, MetaData
+
 
 class SQLiteImpl(DefaultImpl):
     __dialect__ = "sqlite"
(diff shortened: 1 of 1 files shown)


zzzeek:
hi there -

this seems to be doing the "rename table" / "make new table" / "copy data" approach which is already what [batch mode](https://alembic.sqlalchemy.org/en/latest/batch.html) does.  this is already a covered use case, and as long as SQLite itself will not support simple "ALTER" statements, all of that complexity belongs in batch mode.

vvvrrooomm:
ah, great, this PR is so shaky getting things right would have been so much work! 
Is there a chance to promote batch-mode more aggressively.
Something like the last commit 61d116a?

zzzeek:
> ah, great, this PR is so shaky getting things right would have been so much work!
> Is there a chance to promote batch-mode more aggressively.
> Something like the last commit [61d116a](https://github.com/sqlalchemy/alembic/commit/61d116a0e1ecf10690876759aebbc0babdfc4cd7)?

absolutely, SQLite dialect's error messages should be referring to the batch feature which was added later.  sorry if you had missed this!

alembic/ddl/sqlite.py
```diff
@@ -33,17 +33,21 @@ def add_constraint(self, const):
         if const._create_rule is None:
             raise NotImplementedError(
                 "No support for ALTER of constraints in SQLite dialect"
+                "Try to enable batch-mode during auto-migration"
```
zzzeek: Let's make it like:

    Please refer to the batch mode feature which allows for SQLite migrations using a copy-and-move strategy.

vvvrrooomm: ok, done

zzzeek:
probably needs pep8 formatting and also the commit message needs to reflect the current state of things.    will run it through gerrit.

zzzeek:
huh...i have no idea why this is like this, I guess I was trying to get it in review

sqla-tester:
New Gerrit review created for change 5d89cf4e77e5568887f2c8926a589c55427dab12: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1167

sqla-tester:

Code-Review+2 Workflow+1

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/1167

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/1167 has been **merged**. Congratulations! :)

PR #530, opened 2019-01-23
Fix single quote escaping for the SQL comments

### LINKED ISSUE #529: the autogenerate renderer should escape the quotes in the SQL comments
Hi,

I use recent versions of SQLAlchemy and alembic in order to insert some comments on columns and tables (cf. https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/932/).

When I add a column with a comment to a table in my SQLAlchemy schema, the autogenerate feature fails to load the Python migration script due to a single quote (in my comment). I've got a syntax error since the result of the renderer is malformed.

```python
artist = Table(
    "artist",
    metadata,
    Column("name", String, nullable=False, comment="artist name"),
    # added this column with a single quote in the comment
    Column("band", String, comment="Artist's band")
)
```

The `_render_column` in the `autogenerate.render` module returns something like `comment='Artist's band'`. There are three single quotes.

I think the renderer should escape the single quotes, e.g. here https://github.com/sqlalchemy/alembic/blob/master/alembic/autogenerate/render.py#L617

I can replace the `"'%s'"` by a `'"%s"'` or find a proper way to escape the single quote. There might be other lines in the code where the single quote in comment should be escaped.

In my case, I write the comments in French. So I've a lot of *apostrophes*! :slightly_smiling_face: 

I'll submit a PR. I would like to share with you the best way to fix that.

Thanks,
Damien

garaud:
close #529

### DIFF
--- a/tests/test_autogen_render.py
+++ b/tests/test_autogen_render.py
@@ -1155,6 +1155,19 @@ class AutogenRenderTest(TestBase):
             "comment='This is a comment')",
         )
 
+    @config.requirements.comments_api
+    def test_render_col_comment_with_quote(self):
+        c = Column("some_key", Integer, comment="This is a john's comment")
+        Table("some_table", MetaData(), c)
+        result = autogenerate.render._render_column(c, self.autogen_context)
+        print(result)
+        eq_ignore_whitespace(
+            result,
+            "sa.Column('some_key', sa.Integer(), "
+            "nullable=True, "
+            "comment=\"This is a john's comment\")",
+        )
+
     def test_render_col_autoinc_false_mysql(self):
         c = Column("some_key", Integer, primary_key=True, autoincrement=False)
         Table("some_table", MetaData(), c)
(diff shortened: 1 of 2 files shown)


sqla-tester:
New Gerrit review created for change c85f2de986da6c1231a18732b097ff890e67977e: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1111

sqla-tester:

Verified+1 Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1111 has been **merged**. Congratulations! :)

tests/test_autogen_render.py
```diff
@@ -1155,6 +1155,19 @@ def test_render_col_with_comment(self):
             "comment='This is a comment')",
         )
 
+    @config.requirements.comments_api
+    def test_render_col_comment_with_quote(self):
+        c = Column("some_key", Integer, comment="This is a john's comment")
+        Table("some_table", MetaData(), c)
+        result = autogenerate.render._render_column(c, self.autogen_context)
+        print(result)
```
garaud: sorry, there is a remaining print here

PR #537, opened 2019-02-15
Update documentation for Config

### LINKED ISSUE #536: Wrong key used in example in documentation for Config object
In [the documentation on the `Config` class](https://alembic.sqlalchemy.org/en/latest/api/config.html#alembic.config.Config), an example is shown which contains the following code:
```python
from alembic.config import Config
alembic_cfg = Config()
alembic_cfg.set_main_option("script_location", "myapp:migrations")
alembic_cfg.set_main_option("url", "postgresql://foo/bar")
alembic_cfg.set_section_option("mysection", "foo", "bar")
```

Specifically, the line using `"url"` as key should be changed to `"sqlalchemy.url"`, since that is the key that is read later on. Using `"url"` only yields in `KeyError: 'url'` being thrown when the `alembic` command is used.

federicotdn:
Replace "url" with "sqlalchemy.url" in example use of Config object. Fixes #536.

### DIFF
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -40,7 +40,7 @@ class Config(object):
         from alembic.config import Config
         alembic_cfg = Config()
         alembic_cfg.set_main_option("script_location", "myapp:migrations")
-        alembic_cfg.set_main_option("url", "postgresql://foo/bar")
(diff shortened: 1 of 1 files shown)


zzzeek:
merged in 4fb14eeca8dee9d7ec3b2ef2db2d11a329e84abe thanks!

PR #560, opened 2019-05-13
Add database freshness check cookbook recipe

### LINKED ISSUE #559: Cookbook recipe for database freshness check
This feature request is for a cookbook recipe to programatically check database freshness.

Snippet:
```python
from alembic import config
from alembic import script
from alembic.runtime import migration
import sqlalchemy

import exceptions


engine = sqlalchemy.create_engine(DATABASE_URL)
alembic_cfg = config.Config('alembic.ini')
directory = script.ScriptDirectory.from_config(alembic_cfg)
with engine.begin() as conn:
    context = migration.MigrationContext.configure(conn)
    if context.get_current_revision() != directory.get_current_head():
        raise exceptions.DatabaseIsNotUpToDate('Upgrade the database.')
```

(Gist: https://gist.github.com/m-aciek/118d450ee59a41176214b5f93a02cc6f.)

m-aciek:
Closes #559.

Unfortunately I didn't build documentation locally.

### DIFF
--- a/docs/build/cookbook.rst
+++ b/docs/build/cookbook.rst
@@ -1047,3 +1047,28 @@ Output::
     op.create_index('property_name_users_id', 'user_properties', ['property_name', 'users_id'], unique=True)
     # ### end Alembic commands ###
 
+Database freshness check
+========================
+
+Following code answers a question if a database schema is up to date in terms
+of applying Alembic migrations. You may want to check it before running your
+application to raise an exception or log an error.
+
+    from alembic import config, script
+    from alembic.runtime import migration
+    from sqlalchemy import engine
+
+
+    connectable: engine.Connectable
+    alembic_cfg: config.Config
+    
+    directory = script.ScriptDirectory.from_config(alembic_cfg)
+    with connectable.begin() as connection:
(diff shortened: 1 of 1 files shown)


docs/build/cookbook.rst
```diff
@@ -1047,3 +1047,28 @@ Output::
...
+    from sqlalchemy import engine
+
+
+    connectable: engine.Connectable
+    alembic_cfg: config.Config
+    
+    directory = script.ScriptDirectory.from_config(alembic_cfg)
+    with connectable.begin() as connection:
+        context = migration.MigrationContext.configure(connection)
+        return context.get_current_revision() == directory.get_current_head()
```
zzzeek: this needs work because the recipe does not work with branches, and the docs for get_current_revision() state:

> This method intends to be used only for a migration stream that
        does not contain unmerged branches in the target database;
        if there are multiple branches present, an exception is raised.
        The :meth:`.MigrationContext.get_current_heads` should be preferred
        over this method going forward in order to be compatible with
        branch migration support.

will fix

zzzeek:
thanks!

PR #561, opened 2019-05-14
Fix snippet for non-cli usage in empty migrations recipe.

m-aciek:
Now it raises AttributeError when invoked from code.

### DIFF
--- a/docs/build/cookbook.rst
+++ b/docs/build/cookbook.rst
@@ -791,7 +791,7 @@ any operations::
         # ...
 
         def process_revision_directives(context, revision, directives):
-            if config.cmd_opts.autogenerate:
+            if context.opts['revision_context'].command_args['autogenerate']:
                 script = directives[0]
                 if script.upgrade_ops.is_empty():
                     directives[:] = []


docs/build/cookbook.rst
```diff
@@ -791,7 +791,7 @@ any operations::
         # ...
 
         def process_revision_directives(context, revision, directives):
-            if config.cmd_opts.autogenerate:
```
zzzeek: config is present in env.py:

https://github.com/sqlalchemy/alembic/blob/master/alembic/templates/generic/env.py#L11

additionally this could at least be context.config

why are you getting an error?

m-aciek: I am getting error while trying to create revision programaticaly, not via CLI. `cmd_opts` are `None` then. I will try tommorow (CET) `context.config` with my use case.

zzzeek: so the recipe here is given as one someone uses in their env.py, I'm not sure how you'd use it outside of env.py but we can clarify that

zzzeek:
this seems specific to an environment that's not our default so closing.

PR #582, opened 2019-07-03
Allow overriding of script.py.mako path

### LINKED ISSUE #581: Allow overriding of script.py.mako path in generate_revision
We have a use case where our `script.py.mako` is stored in a central location to be shared by multiple sets of alembic revisions. We currently just copy it into each project's scripts directory but that's not ideal.

It looks like this could be allowed by adding an optional kw arg to [generate_revision](https://github.com/sqlalchemy/alembic/blob/930345aec82e7bbd6700ee2c2128116956a38e8e/alembic/script/base.py#L521) which is then passed into `_generate_template`, using the existing value as a default.

gricey432:
Fixes #581

### DIFF
--- a/alembic/script/base.py
+++ b/alembic/script/base.py
@@ -528,6 +528,7 @@ class ScriptDirectory(object):
         branch_labels=None,
         version_path=None,
         depends_on=None,
+        template_src_path=None,
         **kw
     ):
         """Generate a new revision file.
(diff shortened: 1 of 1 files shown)


zzzeek:
hi there!

how does one use this?   env.py doesn't call upon generate_template() directly, things like this are usually set up from [EnvironmentContext.configure()](https://alembic.sqlalchemy.org/en/latest/api/runtime.html?highlight=environmentcontext%20configure#alembic.runtime.environment.EnvironmentContext.configure).  Also all PRs need tests that exercise the feature, in this case we'd be looking probably in test_environment.py.  thanks!

gricey432:
Hi!

Looking back on this a few hours later I'm not really sure what I was trying to do here, you're right it's not even called by anything. I'll close this in its current state

PR #587, opened 2019-07-16
setup.py: set an upper bound on dependencies version

JordanP:
This should make this library more robust in case of breaking changes in one of
its dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,10 +19,10 @@ v.close()
 readme = os.path.join(os.path.dirname(__file__), "README.rst")
 
 requires = [
-    "SQLAlchemy>=0.9.0",
-    "Mako",
-    "python-editor>=0.3",
-    "python-dateutil",
+    "SQLAlchemy>=0.9.0<2.0",
+    "Mako<1.0",
+    "python-editor>=0.3,<2.0",
+    "python-dateutil<3.0",
 ]
 
 


zzzeek:
hi there -

I don't actually agree with these changes at least for Mako, python-editor and python-dateutil.  These are extremely stable packages and are more likely to have bug fixes or new features that an environment would wish to take advantage of, rather than some breaking API change that Alembic needs to guard against.   Alembic's use of python-edtior and python-dateutil are extremely minimal in any case and are very unlikely to cause issues.   Mako has also been beyond version 1.0 for five years and as I'm the maintainer of it, I can guarantee it's API is also not changing in any way that would break on Alembic, so if someone has a particular version of Alembic, I would not want it to be holding back their use of newer versions of any of these packages, which would be my bigger concern.   These upper bounds start becoming problematic in codebases with lots of dependencies as some packages rely on newer versions of things and others have too aggressively limited the upper bound, causing conflicts.

For SQLAlchemy, it's probably more appropriate to have upper bounds here as an older Alembic won't work with newer SQLAlchemy versions.   Current Alembic releases won't work with 1.4 though, but by the time I am making another Alembic release, then 1.4 *will* work because it will also include support for those API changes.    We can put "<2.0" in there, so if someone has an old Alembic it won't work with 2.0, but then, that old Alembic might just as well not work with 1.5 either so is not very useful.    So maybe an upper bound for SQLAlchemy but it likely should start at "< 1.4", then "< 1.5", something like that.

JordanP:
Hi,
Yeah, I completely understand your point. Having also worked for OpenStack for some time, I know that dependencies management is challenging, without universal truth.

I am fine with not doing anything here, having your thoughts and knowing you gave some is already good for me. Though I am not sure I agree with "These are extremely stable packages..." they have been stable for a long time, yep, but maybe that's why I major version bump is likely to be a big thing and break hard.

Anyway, thanks for the detailed answer !

zzzeek:
sure but alembic is using one line from each of these packages, it's almost a philosophical question if one can really know what a package will be after a certain major version bump, or what.     Looking at their recent history, Alembic has moved successfully through dateutil 2.5, 2.6, 2.7, 2.8, etc., I'd hate that they do a 3.0 that doesn't change anything and now everyone has to upgrade their alembic, which might mean they have to upgrade their SQLAlchemy, which might not be easy or possible.       Again, an upper bound on SQLAlchemy itself is likely appropriate because there are definitely breaking changes between that combination.

zzzeek:
I'm doing a bump for lower SQLAlchemy version now.  upper bounds in the package however still make me a little nervous, so im going to pass on this, thanks for the discussion!

Issue #590, opened 2019-07-30
Template in .ini for ScriptDirectory.from_config doesn't translate to windows paths


bubthegreat:
When using the %(here)s/dir format for a script location, the expectation is that alembic should translate this pathing to an OS friendly format for me if needed.  It does not.

Example - in alembic.ini I have the following script_location:
```
[alembic]
# path to migration scripts
script_location = %(here)s/alembic
```
With this code pulling the config:
```
from alembic.config import Config as AlembicConfig
from alembic.script import ScriptDirectory

config_file = 'C:\\Users\\bubth\\Development\\PykuMUD\\src\\pykumud\\alembic.ini'
config = AlembicConfig(config_file)
script = ScriptDirectory.from_config(config)
```
And this directory structure:
```
'C:\\Users\\bubth\\Development\\PykuMUD\\src\\pykumud\\alembic\\env.py',
'C:\\Users\\bubth\\Development\\PykuMUD\\src\\pykumud\\alembic\\script.py.mako',
 'C:\\Users\\bubth\\Development\\PykuMUD\\src\\pykumud\\alembic\\versions',
```
Expectation:
`C:\\Users\\bubth\\Development\\PykuMUD\\src\\pykumud\\alembic`

Result:
`C:\\Users\\bubth\\Development\\PykuMUD\\src\\pykumud/alembic`

bubthegreat:
It seems like the simplest solution to this is to implement os.path.normpath() on any paths we pull from the .ini file.

zzzeek:
OK, I don't understand this issue.    when you say, "Expectation:", "Result:", where is that "result" coming from ?  I don't see you calling upon any API feature that would make use of this path without it being normalized.    When the ScriptDirectory makes use of this path, it uses the _version_locations attribute exclusively, which runs all paths through os.path.abspath(), which [implies normpath](https://docs.python.org/2/library/os.path.html#os.path.abspath).  So I don't know where you are seeing this non-normalized path coming from.

Need information on how to reproduce, thanks!

sqla-tester:
**Mike Bayer** referenced this issue:

**Add test for normalization of script_dir, version_locations** https://gerrit.sqlalchemy.org/1470

zzzeek:
OK I don't have enough info to make any change here.

PR #591, opened 2019-07-30
Fix for mixed os directory path names being parsed properly from .ini file

bubthegreat:
Fix for https://github.com/sqlalchemy/alembic/issues/590 to sanitize paths before we assume they exist - so when on windows, the same .ini can be used as would function on linux when parsing a path from the .ini file.

### DIFF
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -285,7 +285,13 @@ class Config(object):
         indicate a different section.
 
         """
-        return self.get_section_option(self.config_ini_section, name, default)
+        raw_section_option = self.get_section_option(
+            self.config_ini_section, name, default)
+        if name in ('script_location', 'version_locations'):
+            section_option = os.path.normpath(raw_section_option)
+        else:
+            section_option = raw_section_option
+        return section_option
 
 
 class CommandLine(object):


alembic/config.py
```diff
@@ -285,7 +285,13 @@ def get_main_option(self, name, default=None):
         indicate a different section.
 
         """
-        return self.get_section_option(self.config_ini_section, name, default)
+        raw_section_option = self.get_section_option(
```
zzzeek: this logic would take place in get_section_option(), since the "--name" option to alembic can make any section be used.   however I'm not sure this is the most appropriate place for the normpath to occur in any case.

There is already path normalization occurring in alembic/script/base.py such as in https://github.com/sqlalchemy/alembic/blob/master/alembic/script/base.py#L590 , so I dont think normalization is needed for version_locations, it needs to be in https://github.com/sqlalchemy/alembic/blob/master/alembic/script/base.py#L136 for script_location.

alembic/config.py
```diff
@@ -285,7 +285,13 @@ def get_main_option(self, name, default=None):
...
+            self.config_ini_section, name, default)
+        if name in ('script_location', 'version_locations'):
```
zzzeek: version_locations is a comma separated list.

zzzeek:
closing this for now as I need clarification on #590 so that I know what exactly should be tested.

Issue #601, opened 2019-09-24
alembic 1.2.0 broke the alembic.command.stamp() revision parameter


toabctl:
Trying to use OpenStack services with alembic 1.2.0 (currently 1.1.0 is used) is breaking a lot of services with (Ironic example):

```
[...]
    b'  File "ronic/ironic/db/sqlalchemy/migration.py", line 99, in stamp'
    b'    return alembic.command.stamp(config, revision=revision)'
    b"TypeError: stamp() got an unexpected keyword argument 'revision'"
```

Looks like the `stamp` method does no longer accept the parameter `revision` (done in commit fdcc5bf973fec758eaf67cf8e8ad48985caffb65).

I wonder if it would be possible (until 2.0.0 comes up) to support `revision` and `revisions` as parameter to be backward compatible?

Also oslo.db is affected. It tries to use the `revision` parameter. See https://opendev.org/openstack/oslo.db/src/branch/master/oslo_db/sqlalchemy/migration_cli/ext_alembic.py#L92

zzzeek:
someone actually noted that I changed this name, and I asked, you're not actualy using the positional argument as a kw are you, and they said no, and I was like, OK, great.   But then here we are.   Alembic is actually tested against neutron and oslo.db, I can add more alembic-using projects like ironic since nova/keystone have unfortuantely never switched.

There's a decorator I can use to allow for both kw args but also it should be passed positionally.

sqla-tester:
**Thomas Bechtold** has proposed a fix for this issue in the **master** branch:

**Fix alembic.command.stamp() signature and allow revision again** https://gerrit.sqlalchemy.org/1477

zzzeek:
also why didn't oslo.db's test fail in my own CI

zzzeek:
also as far as 1.2 etc we don't use pure semver, see https://www.sqlalchemy.org/download.html#versions.  there's never going to be an alembic 2.0

PR #602, opened 2019-09-24
Fix alembic.command.stamp() signature and allow revision again

toabctl:
Commit [1] changed the method signature. "revision" as parameter name
was no longer accepted. This is now fixed.

Fixes: #601

[1] https://github.com/sqlalchemy/alembic/commit/fdcc5bf973fec758eaf6

### DIFF
--- a/alembic/command.py
+++ b/alembic/command.py
@@ -515,13 +515,13 @@ def current(config, verbose=False, head_only=False):
         script.run_env()
 
 
-def stamp(config, revisions, sql=False, tag=None, purge=False):
+def stamp(config, revision, sql=False, tag=None, purge=False):
     """'stamp' the revision table with the given revision; don't
     run any migrations.
 
     :param config: a :class:`.Config` instance.
 
-    :param revision: target revision.
+    :param revision: target revision. Can be a list of revisions.
 
     :param sql: use ``--sql`` mode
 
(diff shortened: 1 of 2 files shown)


zzzeek:
hi there -

thanks for this but we have a "legacy names" decorator that will allow the old name with a deprecation warning.

zzzeek:
which doesnt work here.

sqla-tester:
New Gerrit review created for change 4126635e272f82ec53d630bb77c00cb6ff42cec9: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1477

zzzeek:
the problem is this breaks the CLI.  the name "revisions" means argparse accepts a list of arguments, "revision" means it does not.    There is no way for this to be fixed without a dramatic change to how the whole system works, not like we shouldn't be using click or something at this point, but this is just a single name change.     I'm considering that openstack might just have to push their code upwards to use 1.2, or just hardcoding a lot in the config system to work around this, but most ideal would be some way that the name can change but I don't see a way for that to be possible without changing the function to use `*args` or something awful like that.

sqla-tester:
**Thomas Bechtold** (toabctl) wrote:

Code-Review+1

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1477 has been **merged**. Congratulations! :)

PR #607, opened 2019-10-03
Configuration file from environment variable

p1c2u:
Adds support for setting configuration file from `ALEMBIC_CONFIG` environment variable

### DIFF
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -61,6 +61,10 @@ class Config(object):
     :param file\_: name of the .ini file to open.
     :param ini_section: name of the main Alembic section within the
      .ini file
+    :param environ: environment variables dictionary.
+     Defaults to ``os.environ``.
+    :param environ_var: name of the environment variable with .ini file path.
+     Defaults to ``ALEMBIC_CONFIG``.
     :param output_buffer: optional file-like input buffer which
      will be passed to the :class:`.MigrationContext` - used to redirect
      the output of "offline generation" when using Alembic programmatically.
@@ -94,6 +98,8 @@ class Config(object):
         self,
         file_=None,
         ini_section="alembic",
+        environ=os.environ,
+        environ_var="ALEMBIC_CONFIG",
         output_buffer=None,
         stdout=sys.stdout,
         cmd_opts=None,
@@ -103,8 +109,10 @@ class Config(object):
         """Construct a new :class:`.Config`
 
         """
-        self.config_file_name = file_
+        self.config_file = file_
         self.config_ini_section = ini_section
+        self.config_environ = environ
+        self.config_environ_var = environ_var
         self.output_buffer = output_buffer
         self.stdout = stdout
         self.cmd_opts = cmd_opts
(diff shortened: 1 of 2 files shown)


zzzeek:
hi there - 

This is not stated in the PR guidelines for alembic but I very much prefer that an issue is opened first for a new feature so that the rationale may be discussed.   That way you don't have to go through all the trouble of writing a PR with tests (looks great btw except it would need documentation) in case we decide this is not how we want to actually solve this problem.

in the test I see some pushing of values into config but what's the full rationale for this feature ?    

I assume what you are going for is to impact the command-line environment first and foremost:

    alembic -c <config_file>

becomes:

  ALEMBIC_CONFIG=<config_file> alembic

is that it?     Is there a reason this is not sufficient:

   alembic -c ${ALEMBIC_CONFIG}

p1c2u:
Hi @zzzeek . Thank you for your comment. I created issue for the PR as you suggested.

sqla-tester:
New Gerrit review created for change 0ea7f8fea69f2be5f9ce5be2343149bf62f0dfcb: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1502

zzzeek:
IIUC this will not work if ALEMBIC_CONFIG is set and then someone passes -c as well, -c should supersede the environment variable

alembic/config.py
```diff
@@ -103,8 +109,10 @@ def __init__(
         """Construct a new :class:`.Config`
 
         """
-        self.config_file_name = file_
```
zzzeek: this is all very complicated, and won't work if someone specifies -c, why can't isnt ALEMBIC_CONFIG simply consumed inside of CommandLine.main() when we pass the name of the config file to the Config (one line change basically?)  is there some reason you need programmatic access to os.environ within the Config object itself  as well as the ability to make use of an alternate dictionary to os.environ ?

zzzeek:
does this patch work for you ?

```diff
diff --git a/alembic/config.py b/alembic/config.py
index b0700d5..22d8152 100644
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -468,12 +468,14 @@ class CommandLine(object):
                     subparser.add_argument(arg, help=positional_help.get(arg))
 
         parser = ArgumentParser(prog=prog)
+
         parser.add_argument(
             "-c",
             "--config",
             type=str,
-            default="alembic.ini",
-            help="Alternate config file",
+            default=os.environ.get("ALEMBIC_CONFIG", "alembic.ini"),
+            help="Alternate config file; defaults to value of "
+            "ALEMBIC_CONFIG environment variable, or \"alembic.ini\"",
         )
         parser.add_argument(
             "-n",

```

zzzeek:
OK I have this proposed in  https://gerrit.sqlalchemy.org/1514

sqla-tester:

going with  https://gerrit.sqlalchemy.org/1514 for now

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1502 has been **abandoned**.  That means that at least for the moment I need to close this pull request. Sorry it didn't work out :(

Issue #605, opened 2019-10-01
MySQL Boolean autogenerate comparson to TINYINT


johaven:
Alembic version: 1.2.1 
SQL: MariaDB

Each upgrade  (with compare_type enabled) duplicate constraints :

```python
paid = db.Column(db.Boolean)
satisfaction = db.Column(db.Boolean)
```

```sql
  CONSTRAINT `CONSTRAINT_1` CHECK (`paid` in (0,1)),
  CONSTRAINT `CONSTRAINT_2` CHECK (`satisfaction` in (0,1)),
  CONSTRAINT `CONSTRAINT_3` CHECK (`paid` in (0,1)),
  CONSTRAINT `CONSTRAINT_4` CHECK (`satisfaction` in (0,1)),
  CONSTRAINT `CONSTRAINT_5` CHECK (`paid` in (0,1)),
  CONSTRAINT `CONSTRAINT_6` CHECK (`satisfaction` in (0,1)),
  CONSTRAINT `CONSTRAINT_7` CHECK (`paid` in (0,1)),
  CONSTRAINT `CONSTRAINT_8` CHECK (`satisfaction` in (0,1)),
  CONSTRAINT `CONSTRAINT_9` CHECK (`paid` in (0,1)),
  CONSTRAINT `CONSTRAINT_10` CHECK (`satisfaction` in (0,1)),
  CONSTRAINT `CONSTRAINT_11` CHECK (`paid` in (0,1)),
  CONSTRAINT `CONSTRAINT_12` CHECK (`satisfaction` in (0,1)),
  CONSTRAINT `CONSTRAINT_13` CHECK (`paid` in (0,1)),
  CONSTRAINT `CONSTRAINT_14` CHECK (`satisfaction` in (0,1)),
  CONSTRAINT `CONSTRAINT_15` CHECK (`paid` in (0,1)),
  CONSTRAINT `CONSTRAINT_16` CHECK (`satisfaction` in (0,1)),
  CONSTRAINT `CONSTRAINT_17` CHECK (`paid` in (0,1)),
  CONSTRAINT `CONSTRAINT_18` CHECK (`satisfaction` in (0,1))
```

zzzeek:
this is because compare_type does not detect a MySQL TINYINT column as the same as Boolean, so you will need to adjust these migrations manually or use a custom type comparator.

there is then a second issue that when "changing" the type of the column, it does not drop the old constraint.

[...]

PR #619, opened 2019-11-08
Update the type comparison code used for schema autogeneration. Compare

pbecotte:
the output text for the type to look for changes. In addition, allow
schemas to define sets of types that are functionally equivalent, such
as BOOL and TINYINT(1).
#605

### DIFF
--- a/alembic/ddl/impl.py
+++ b/alembic/ddl/impl.py
@@ -1,6 +1,8 @@
+from collections import namedtuple
+import re
+
 from sqlalchemy import schema
 from sqlalchemy import text
-from sqlalchemy import types as sqltypes
 
 from . import base
 from .. import util
@@ -21,6 +23,9 @@ class ImplMeta(type):
 _impls = {}
 
 
+Params = namedtuple("Params", ["type", "args", "kwargs"])
+
+
 class DefaultImpl(with_metaclass(ImplMeta)):
 
     """Provide the entrypoint for major migration operations,
@@ -39,6 +44,7 @@ class DefaultImpl(with_metaclass(ImplMeta)):
 
     transactional_ddl = False
     command_terminator = ";"
+    type_synonyms = ({"NUMERIC", "DECIMAL"},)
 
     def __init__(
         self,
@@ -322,32 +328,62 @@ class DefaultImpl(with_metaclass(ImplMeta)):
                     for row in rows:
                         self._exec(table.insert(inline=True).values(**row))
 
-    def compare_type(self, inspector_column, metadata_column):
+    def _tokenize_column_type(self, column):
+        definition = self.dialect.type_compiler.process(column.type).lower()
+        # py27 - py36 - col_type, *param_terms = re.findall...
+        matches = re.findall("[^(),]+", definition)
+        col_type, param_terms = matches[0], matches[1:]
+        params = Params(col_type, [], {})
+        for term in param_terms:
+            if "=" in term:
+                key, val = term.split("=")
+                params.kwargs[key.strip()] = val.strip()
+            else:
+                params.args.append(term.strip())
+        return params
 
-        conn_type = inspector_column.type
-        metadata_type = metadata_column.type
-
-        metadata_impl = metadata_type.dialect_impl(self.dialect)
-        if isinstance(metadata_impl, sqltypes.Variant):
-            metadata_impl = metadata_impl.impl.dialect_impl(self.dialect)
-
-        # work around SQLAlchemy bug "stale value for type affinity"
-        # fixed in 0.7.4
-        metadata_impl.__dict__.pop("_type_affinity", None)
-
-        if hasattr(metadata_impl, "compare_against_backend"):
-            comparison = metadata_impl.compare_against_backend(
-                self.dialect, conn_type
-            )
-            if comparison is not None:
-                return not comparison
-
-        if conn_type._compare_type_affinity(metadata_impl):
-            comparator = _type_comparators.get(conn_type._type_affinity, None)
-
-            return comparator and comparator(metadata_impl, conn_type)
-        else:
+    def _column_types_match(self, inspector_type, metadata_type):
+        if inspector_type == metadata_type:
             return True
+        synonyms = [{t.lower() for t in batch} for batch in self.type_synonyms]
+        for batch in synonyms:
+            if {inspector_type, metadata_type}.issubset(batch):
+                return True
+        return False
+
+    def _column_args_match(self, inspected_params, meta_params):
+        """We want to compare column parameters. However, we only want
+        to compare parameters that are set. If they both have `collation`,
+        we want to make sure they are the same. However, if only one
+        specifies it, dont flag it for being less specific
+        """
+        # py27- .keys is a set in py36
+        for kw in set(inspected_params.kwargs.keys()) & set(
+            meta_params.kwargs.keys()
+        ):
+            if inspected_params.kwargs[kw] != meta_params.kwargs[kw]:
+                return False
+
+        for meta, inspect in zip(inspected_params.args, meta_params.args):
+            if meta != inspect:
+                return False
+
+        return True
+
+    def compare_type(self, inspector_column, metadata_column):
+        """Returns True if there ARE differences between the types of the two
+        columns. Takes impl.type_synonyms into account between retrospected
+        and metadata types
+        """
+        inspector_params = self._tokenize_column_type(inspector_column)
+        metadata_params = self._tokenize_column_type(metadata_column)
+        if not self._column_types_match(
+            inspector_params.type, metadata_params.type
+        ):
+            return True
+        if not self._column_args_match(inspector_params, metadata_params):
+            return True
+        return False
 
     def compare_server_default(
         self,
@@ -425,45 +461,3 @@ class DefaultImpl(with_metaclass(ImplMeta)):
 
     def render_type(self, type_obj, autogen_context):
         return False
-
-
-def _string_compare(t1, t2):
-    return t1.length is not None and t1.length != t2.length
-
-
-def _numeric_compare(t1, t2):
-    return (t1.precision is not None and t1.precision != t2.precision) or (
-        t1.precision is not None
-        and t1.scale is not None
-        and t1.scale != t2.scale
-    )
-
-
-def _integer_compare(t1, t2):
-    t1_small_or_big = (
-        "S"
-        if isinstance(t1, sqltypes.SmallInteger)
-        else "B"
-        if isinstance(t1, sqltypes.BigInteger)
-        else "I"
-    )
-    t2_small_or_big = (
-        "S"
-        if isinstance(t2, sqltypes.SmallInteger)
-        else "B"
-        if isinstance(t2, sqltypes.BigInteger)
-        else "I"
-    )
-    return t1_small_or_big != t2_small_or_big
-
-
-def _datetime_compare(t1, t2):
-    return t1.timezone != t2.timezone
-
-
-_type_comparators = {
-    sqltypes.String: _string_compare,
-    sqltypes.Numeric: _numeric_compare,
-    sqltypes.Integer: _integer_compare,
-    sqltypes.DateTime: _datetime_compare,
-}
--- a/docs/build/autogenerate.rst
+++ b/docs/build/autogenerate.rst
@@ -132,10 +132,15 @@ Autogenerate can **optionally detect**:
 * Change of column type.  This will occur if you set
   the :paramref:`.EnvironmentContext.configure.compare_type` parameter
   to ``True``, or to a custom callable function.   The default implementation
-  **only detects major type changes**, such as between ``Numeric`` and
-  ``String``, and does not detect changes in arguments such as lengths, precisions,
-  or enumeration members.  The type comparison logic is extensible to work
-  around these limitations, see :ref:`compare_types` for details.
+  will reliably detect major changes, such as between ``Numeric`` and
+  ``String``.  If both the existing type and the new type have keyword arguments
+  specified (such as lengths, precisions, or enumeration members), they will be
+  compared as well.  However, if keywords are only specified for one,
+  changes in these will be ignored. The type comparison logic is extensible to
+  work around these limitations, see :ref:`compare_types` for details.
+
+.. versionchanged:: 1.4 type comparison code has been enhanced to compare column types more deeply as well as to take arguments into account.
+
 * Change of server default.  This will occur if you set
   the :paramref:`.EnvironmentContext.configure.compare_server_default`
   parameter to ``True``, or to a custom callable function.
@@ -405,18 +410,26 @@ is set to True::
 .. note::
 
    The default type comparison logic (which is end-user extensible) currently
-   works for **major changes in type only**, such as between ``Numeric`` and
-   ``String``.     The logic will **not** detect changes such as:
+   (as of Alembic version 1.4.0) works by comparing the generated SQL for a
+   column. It does this in two steps-
 
-   * changes between types that have the same "type affinity", such as
-     between ``VARCHAR`` and ``TEXT``, or ``FLOAT`` and ``NUMERIC``
+   * First, it compares the outer type of each column such as ``VARCHAR``
+     or ``TEXT``. Dialect implementations can have synonyms that are considered
+     equivalent- this is because some databases support types by converting them
+     to another type. For example, NUMERIC and DECIMAL are considered equivalent
+     on all backends, while on the Oracle backend the additional synonyms
+     BIGINT, INTEGER, NUMBER, SMALLINT are added to this list of equivalents
 
-   * changes between the arguments within the type, such as the lengths of
+   * Next, the arguments within the type, such as the lengths of
      strings, precision values for numerics, the elements inside of an
-     enumeration.
+     enumeration are compared. If BOTH columns have arguments AND they are
+     different, a change will be detected. If one column is just set to the
+     default and the other has arguments, Alembic will pass on attempting to
+     compare these. The rationale is that it is difficult to detect what a
+     database backend sets as a default value without generating false
+     positives.
 
-   Detection of these kinds of parameters is a long term project on the
-   SQLAlchemy side.
+..versionchanged 1.4.0:: Added the text and keyword comparison for column types
 
 Alternatively, the :paramref:`.EnvironmentContext.configure.compare_type`
 parameter accepts a callable function which may be used to implement custom type
@@ -487,6 +500,9 @@ then a basic check for type equivalence is run.
 
 .. versionadded:: 0.7.6 - added support for the ``compare_against_backend()``
    method.
+
+.. versionadded:: 1.4.0 - added column keyword comparisons and the
+   ``type_synonyms`` property.
 
 
 .. _post_write_hooks:
--- a/tests/requirements.py
+++ b/tests/requirements.py
@@ -5,6 +5,10 @@ from alembic.util import sqla_compat
 
 class DefaultRequirements(SuiteRequirements):
     @property
+    def unicode_string(self):
+        return exclusions.skip_if(["oracle"])
+
+    @property
     def alter_column(self):
         return exclusions.skip_if(["sqlite"], "no ALTER COLUMN support")
 
@@ -85,6 +89,12 @@ class DefaultRequirements(SuiteRequirements):
         return exclusions.fails_on_everything_except(
             "postgresql", "oracle", "mssql", "sybase", "sqlite"
         )
+
+    @property
+    def datetime_timezone(self):
+        """target dialect supports timezone with datetime types."""
+
+        return exclusions.only_on(["postgresql"])
 
     @property
     def postgresql_uuid_ossp(self):
--- a/tests/test_autogen_diffs.py
+++ b/tests/test_autogen_diffs.py
@@ -2,16 +2,21 @@ import sys
 
 from sqlalchemy import BIGINT
 from sqlalchemy import BigInteger
+from sqlalchemy import Boolean
 from sqlalchemy import CHAR
 from sqlalchemy import CheckConstraint
 from sqlalchemy import Column
+from sqlalchemy import DATE
 from sqlalchemy import DateTime
 from sqlalchemy import DECIMAL
+from sqlalchemy import Enum
+from sqlalchemy import FLOAT
 from sqlalchemy import ForeignKey
 from sqlalchemy import ForeignKeyConstraint
 from sqlalchemy import Index
 from sqlalchemy import INTEGER
 from sqlalchemy import Integer
+from sqlalchemy import LargeBinary
 from sqlalchemy import MetaData
 from sqlalchemy import Numeric
 from sqlalchemy import PrimaryKeyConstraint
@@ -20,7 +25,9 @@ from sqlalchemy import String
 from sqlalchemy import Table
 from sqlalchemy import Text
 from sqlalchemy import text
+from sqlalchemy import TIMESTAMP
 from sqlalchemy import TypeDecorator
+from sqlalchemy import Unicode
 from sqlalchemy import UniqueConstraint
 from sqlalchemy import VARCHAR
 from sqlalchemy.dialects import sqlite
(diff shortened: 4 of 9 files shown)


sqla-tester:
New Gerrit review created for change e4aeb65eb47b5ccbb3de61fad3efef83dad730cd: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

zzzeek:
the jenkins runner will run some openstack suites which in the case of neutron might exercise the type comparison stuff a bit.   let's see what happens.

sqla-tester:

Let me start off with, thank you very much for this work so far, it looks amazing!   This is a really critical area where we get a lot of complaints and I am really excited if we can finally make it work for real.

so there's a bunch of failures but they look to be all in the realm of unexpected positives for the comparison logic.

in the main suite, we have for example, an unexpected change in type between VARCHAR and String (this is likely because on Oracle, it's rendered as VARCHAR2 ? )

    File "/home/jenkins/workspace/alembic_gerrit/1d95c17e/tests/test_autogen_comments.py", line 101, in test_alter_table_comment
      eq_(diffs[0][0], "add_table_comment")
    File "/home/jenkins/workspace/alembic_gerrit/1d95c17e/.tox/py37-sqlamaster-sqlite-postgresql-mysql-oracle/lib/python3.7/site-packages/sqlalchemy/testing/assertions.py", line 237, in eq_
      assert a == b, msg or "%r != %r" % (a, b)
  AssertionError: ('modify_type', None, 'some_table', 'test', {'existing_nullable': None, 'existing_server_default': False, 'existing_comment': None}, VARCHAR(length=10), String(length=10)) != 'add_table_comment'
        


over in openstack nova, where they run tests that compare schemas to be the same on MySQL, where the VARCHAR datatype has a lot of flags, we see:

  AssertionError: Models and migration scripts aren't in sync:
  [ [ ( 'modify_type',
        None,
        'instance_type_extra_specs',
        'key',
        { 'existing_comment': None,
          'existing_nullable': True,
          'existing_server_default': False},
        VARCHAR(collation='utf8_bin', length=255),
        String(length=255))],
    [ ( 'modify_type',
        None,
        'instance_type_extra_specs',
        'value',
        { 'existing_comment': None,
          'existing_nullable': True,
          'existing_server_default': False},
        VARCHAR(collation='utf8_bin', length=255),
        String(length=255))],
    [ ( 'modify_type',
        None,
        'resource_providers',
        'name',
        { 'existing_comment': None,
          'existing_nullable': True,
          'existing_server_default': False},
        VARCHAR(charset='utf8', length=200),
        Unicode(length=200))]]

those are all false positives; Unicode / String render VARCHAR on MySQL and there are default values for collation and charset.   I think a good rule here might be if the DB-reflected column has some flags that are set, and the model-side has no flag set for that value at all, that means no change.

so this opens up a new subject area which is that when we add new comparison features to autogenerate, especially when we added indexes/unique constraints, the vast majority of issues reported were related to false positives.    The index/unique thing, I must have had 20 issues over the course of several years with that.     So, we might want to build this so that it's being very conservative about reporting a change based on keyword arguments that are different.   

as always, let me know how you are doing and if you have motivation / resources to keep going.  I can always pick up sooner rather than later if need be but I am glad if you'd like to keep going!

pbecotte:
That is a WAY better idea. Felt pretty ugly adding in all those default keywords into the synonyms list, and I knew I would miss a bunch. But only comparing keywords if they BOTH have them is exactly the right thing to do!

I can add `{"string", "varchar2"}` to the oracle impl... but I don't know how to test against oracle. they don't seem to have an easily used docker container? I haven't touched oracle since about 2012 :)

I pushed up an update. I have been testing with postgresql/mysql/sqlite. I think this version of the code is a little easier to follow as well... but still feels too complicated?

pbecotte:
Oh, question- I made the assumption that `(` is always the split between the type and any keywords- does that hold true across all the databases/types?

zzzeek:
> Oh, question- I made the assumption that `(` is always the split between the type and any keywords- does that hold true across all the databases/types?

as far as I know, sure.   I would think that all the logic here would be overridable by a subclass impl if they needed to get in there and change things.

sqla-tester:
Patchset 94e7ad0a7254ef17f54039179e8ae71199f92aaa added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

pbecotte:
Added the VARCHAR/VARCHAR2 synonym to oracle

sqla-tester:
Patchset 8963755584c95731a46d307fe3cbd5ccf710b2be added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset 8963755584c95731a46d307fe3cbd5ccf710b2be added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset 17a39d600427c0fc97d09470ad53b43e77f95e1e added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset 2edf4641ce393cf1d7286d90eaf10ce2b87ba550 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset 07495dddfc0bf83dfdc1dcdb76e76e3181550f01 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset 34a6c9e20dc0151e68b1cb801fc98fc99ccd4ad1 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset dbb58692a5a9b3ad7f3326d7d11f1f76d2adcb04 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset f07fe9675daccf5f11191f5a66625251b5ede9cd added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset ba07631c2cf9542def945c7f494f11af1285bab3 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

alembic/ddl/impl.py
```diff
@@ -322,32 +328,63 @@ def bulk_insert(self, table, rows, multiinsert=True):
...
 
-        metadata_impl = metadata_type.dialect_impl(self.dialect)
-        if isinstance(metadata_impl, sqltypes.Variant):
-            metadata_impl = metadata_impl.impl.dialect_impl(self.dialect)
+    def _column_types_match(self, inspector_type, metadata_type):
+        if inspector_type == metadata_type:
+            return True
+        synonyms = [{t.lower() for t in batch} for batch in self.type_synonyms]
+        for batch in synonyms:
+            if {inspector_type, metadata_type}.issubset(batch):
```
sqla-tester: **mike bayer** (zzzeek) wrote:

brilliant!

tests/test_autogen_diffs.py
```diff
@@ -705,6 +711,111 @@ def test_integer(self):
...
+
+    def _fixture(self):
+        from alembic.ddl.postgresql import PostgresqlImpl
+        from sqlalchemy.dialects import postgresql
+
+        return PostgresqlImpl(
+            postgresql.dialect(), None, False, True, None, {}
+        )
+
     def test_datetime(self):
```
sqla-tester: **mike bayer** (zzzeek) wrote:

Interesting, that you need the PG dialect now.  that is fine.   However, I think the test that is here should also be in the SQLite version, where we assert that the compare_type is False in all cases.  this gets into the bigger issue of comparing types that have dialect-specific flags and those that use Variants.  if look at AutogenerateVariantCompareTest in this same file we can probably add some more fun tests with things like `String().with_variant(mysql.VARCHAR(collation='xyz'), "mysql")`.

pbecotte: Now that you point it out, probably didn't actually need that here. Wrote this test when the code was much less general. Let me have a look.

pbecotte:
I like the idea of adding more thorough testing with some factory setup. Two questions there- how far should I go with something like that? (a big change to testing code along with a core change to real code may make the changeset as a whole too large. Of course, NOT adding those tests could leave some weirdness around the actual logic that we miss. Maybe I can put in some extra testing stuff in an separate change and then come back to merging this). and 2. If we're thinking about combinatoric style testing, what are your thoughts on [Hypothesis](https://hypothesis.readthedocs.io/en/latest/)?

zzzeek:
for combinatoric testing we are using py.test combinations, which is part of the test framework in SQLAlchemy as of 1.3.  however, it is not present in Alembic which is still supporting SQLAlchemy 1.1 and upwards.   so we would need to port the combinatoric features from SQLAlchemy into Alembic's testing framework, which is a bit of a PITA, or if we just dropped support for older SQLAlchemy.

SQLAlchemy itself has a signfiicant testing framework that predates py.test but now integrates as much as possible with it.    Both projects are very far along / mature and I'd rather not have any hardcoded dependencies in the test suite (made that mistake with nose already).   i don't see it as worthwhile to switch everything to a new framework right now.

zzzeek:
here's some recent combinatoric testing set up:   https://github.com/sqlalchemy/sqlalchemy/commit/240d9a60ccdb540543a72d9ff30a6f50d33acc5d

pbecotte:
Cleaning this up (the combinations stuff is awesome!) I have a conflict I'm not sure how to solve. I remember why I only ran the datetime test on postgres- that was the only db that actually renders datetime with the timezone construct. So on mysql/sqlite both DATETIME(timezone=True) and DATETIME(timezone=False) render as `DATETIME`. 

This makes a test like https://github.com/sqlalchemy/alembic/blob/master/tests/test_autogen_diffs.py#L678
fail with the new code. I am not sure the right solution though- SHOULD adding the timezone argument to a datetime in a db that doesn't support it cause an alter column to be generated? If not, I can fix the tests to expect no changes if the db doesn't support timezones. If they do, I'll have to figure out a way of comparing datetimes that takes that into account.

zzzeek:
> Cleaning this up (the combinations stuff is awesome!) I have a conflict I'm not sure how to solve. I remember why I only ran the datetime test on postgres- that was the only db that actually renders datetime with the timezone construct. So on mysql/sqlite both DATETIME(timezone=True) and DATETIME(timezone=False) render as `DATETIME`.
> 
> This makes a test like https://github.com/sqlalchemy/alembic/blob/master/tests/test_autogen_diffs.py#L678
> fail with the new code. I am not sure the right solution though- SHOULD adding the timezone argument to a datetime in a db that doesn't support it cause an alter column to be generated? If not, I can fix the tests to expect no changes if the db doesn't support timezones. If they do, I'll have to figure out a way of comparing datetimes that takes that into account.

if autogen is run on a database that doesnt support a timezone setting, and that's all that changed, then it should not generate a change for that.

there is a way to mark this specific combination as, "passes only if the database supports timezone", first by adding a requirements rule in https://github.com/sqlalchemy/alembic/blob/master/tests/requirements.py:

```
    @property
    def datetime_timezone(self):
        """target dialect supports timezone with datetime types."""

        return exclusions.succeeds_if(["postgresql"])

```

then you can add it as a rule to the combination as an additional member in the tuple:

`    (DateTime(), DateTime(timezone=False), False, config.requirements.datetime_timezone),`

the combinations wrapper i wrote around pytest also adapts our exclusion rules system (which predates pytest's existence by many years) for specific sets of parameters.

sqla-tester:

where we at on this, is there more to do?

pbecotte:
Am working on the cleanup discussed above, it'll be soon if nothing else comes up

zzzeek:
just making sure you weren't waiting on me

zzzeek:
oh well github has "request a review" so click for me and/or sqla-tester that when it's next ready :)

sqla-tester:
Failed to create a gerrit review, git squash against branch 'master' failed

zzzeek:
do you know how to rebase against master and force-push here?

pbecotte:
Yeah sorry, pushed the button right as I got off the train. Will rebase after work

sqla-tester:
Patchset 2c9faf6ddc97169579b2b4772c9699d1213037a0 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset b80f22532b8677c7778a8601b8da931d83eac775 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

sqla-tester:
Patchset 97da3483f4d30283abe75b403165af276c48d691 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

alembic/testing/plugin/pytestplugin.py
```diff
@@ -50,6 +52,30 @@ def pytest_pycollect_makeitem(collector, name, obj):
         return []
 
 
+def vendored_exclusions(compound_object):
```
pbecotte: As you mentioned, the testing construct isn't in sqlalchemy before 1.3. So I did this janky thing- though I think it may make more sense to just drop the support for older versions as you mentioned.

alembic/ddl/impl.py
```diff
@@ -39,6 +44,7 @@ class DefaultImpl(with_metaclass(ImplMeta)):
 
     transactional_ddl = False
     command_terminator = ";"
+    type_synonyms = [{"NUMERIC", "DECIMAL"}]
```
sqla-tester: **mike bayer** (zzzeek) wrote:

I tend to use tuples for things like this to make sure nothing is mutated...

alembic/testing/plugin/pytestplugin.py
sqla-tester: **mike bayer** (zzzeek) wrote:

I've re-vendored this into alembic again in alembic/testing/exclusions.py so we can remove this

tests/requirements.py
```diff
@@ -5,6 +5,10 @@
 
 
 class DefaultRequirements(SuiteRequirements):
+    @property
```
sqla-tester: **mike bayer** (zzzeek) wrote:

the def name should be stated in terms of whatever it is that oracle is not doing, not the DB name itself

tests/test_autogen_diffs.py
```diff
@@ -39,9 +47,10 @@
 from alembic.testing import is_not_
 from alembic.testing import mock
 from alembic.testing import TestBase
+from alembic.testing.env import clear_staging_env, staging_env
 from alembic.util import CommandError
-from ._autogen_fixtures import AutogenFixtureTest
-from ._autogen_fixtures import AutogenTest
+from ._autogen_fixtures import _default_object_filters, AutogenFixtureTest, \
```
sqla-tester: **mike bayer** (zzzeek) wrote:

this is all going to reformat w/ black...if you install the .pre-commit-config.yml hook for your next commit

pbecotte: is the project switching to black? I had it in black format and undid it in some places to match the older style haha. I approve! will do

sqla-tester:

> (1 comment)
 > 
 > pbecotte@github wrote:

alembic has been on black for many months now, actually the lines here would reformat because of the "zimports" tool I wrote first and foremost.   Just turn on the pre-commit hooks and it will all be taken care of :).

alembic/ddl/impl.py
```diff
@@ -322,32 +328,63 @@ def bulk_insert(self, table, rows, multiinsert=True):
...
+            if inspected_params.kwargs[kw] != meta_params.kwargs[kw]:
+                return False
 
-        if hasattr(metadata_impl, "compare_against_backend"):
-            comparison = metadata_impl.compare_against_backend(
-                self.dialect, conn_type
-            )
-            if comparison is not None:
-                return not comparison
+        for meta, inspect in zip(inspected_params.args, meta_params.args):
```
sqla-tester: **mike bayer** (zzzeek) wrote:

do we know that these two .args are the same length?  What if a VARCHAR has a default length that gets set up on the database side but isn't in the metadata?  i think perhaps if len(meta_params.args) == 0, we skip this test.

docs/build/autogenerate.rst
```diff
@@ -132,10 +132,12 @@ Autogenerate can **optionally detect**:
...
-  **only detects major type changes**, such as between ``Numeric`` and
-  ``String``, and does not detect changes in arguments such as lengths, precisions,
-  or enumeration members.  The type comparison logic is extensible to work
-  around these limitations, see :ref:`compare_types` for details.
+  will reliably detect major changes, such as between ``Numeric`` and
+  ``String``.  If both the existing type and the new type have keyword arguments
+  specified (such as lengths, precisions, or enumeration members), they will be
+  compared as well.  However, if keywords are only specified for one,
+  changes in these will be ignored. The type comparison logic is extensible to
+  work around these limitations, see :ref:`compare_types` for details.
```
sqla-tester: **mike bayer** (zzzeek) wrote:

add:

  .. versionchanged:: 1.4   type comparison code has been enhanced to compare column types more deeply as well as to take arguments into account.

docs/build/autogenerate.rst
```diff
@@ -405,18 +407,24 @@ is set to True::
 .. note::
 
    The default type comparison logic (which is end-user extensible) currently
```
sqla-tester: **mike bayer** (zzzeek) wrote:

"as of Alembic version 1.4"   just to make sure people see this

docs/build/autogenerate.rst
```diff
@@ -405,18 +407,24 @@ is set to True::
...
-     between ``VARCHAR`` and ``TEXT``, or ``FLOAT`` and ``NUMERIC``
 
-   * changes between the arguments within the type, such as the lengths of
-     strings, precision values for numerics, the elements inside of an
-     enumeration.
+   * First, it compares the outer type of each column such as ``VARCHAR``
+     or ``TEXT``. Dialect implementations can have synonyms that are considered
+     equivalent- this is because some databases support types by converting them
+     to another type. For example, in ``NUMERIC`` and ``DECIMAL`` are considered
+     equivalent.
```
sqla-tester: **mike bayer** (zzzeek) wrote:

"for example, NUMERIC and DECIMAL are considered equivalent on all backends, while on the Oracle backend the additional synonyms BIGINT, INTEGER, NUMBER, SMALLINT are added to this list of equivalents"

docs/build/autogenerate.rst
```diff
@@ -405,18 +407,24 @@ is set to True::
...
 
-   Detection of these kinds of parameters is a long term project on the
-   SQLAlchemy side.
+   * Next, the arguments within the type, such as the lengths of
+     strings, precision values for numerics, the elements inside of an
+     enumeration are compared. If BOTH columns have arguments AND they are
+     different, a change will be detected. If one column is jsut set to the
```
sqla-tester: **mike bayer** (zzzeek) wrote:

typo "just"

docs/build/autogenerate.rst
```diff
@@ -405,18 +407,24 @@ is set to True::
...
+     default and the other has arguments, we don't currently detect this. The
```
sqla-tester: **mike bayer** (zzzeek) wrote:

"and the other has arguments, Alembic will pass on attempting to compare these.  The rationale is that it is difficult to detect what a database backend sets as a default value without generating false positives".

docs/build/autogenerate.rst
```diff
@@ -405,18 +407,24 @@ is set to True::
...
+     reason here is that it can be hard to know what the database backend sets
+     as the default values, meaning that INTEGER() and INTEGER(10) could
+     actually be the same thing, and we don't want to regenerate a diff every
+     time.
```
sqla-tester: **mike bayer** (zzzeek) wrote:

add a `.. versionchanged:: 1.4` note here also

docs/build/autogenerate.rst
```diff
@@ -488,6 +496,13 @@ then a basic check for type equivalence is run.
 .. versionadded:: 0.7.6 - added support for the ``compare_against_backend()``
    method.
 
+For a custom dialect, you could also specify ``impl.type_synonyms``. This
```
sqla-tester: **mike bayer** (zzzeek) wrote:

this is getting into people making their own "impls" which I think is out of scope for this document, so I would take this blurb out.

tests/test_autogen_diffs.py
```diff
@@ -693,6 +692,135 @@ def test_compare_type(
...
+    def test_datetime_comparisons(self, cola, colb, expect_changes):
+        is_(self._compare_columns(cola, colb), expect_changes)
+
+    @testing.combinations(
+        (Integer(), Integer(), False),
+        (
+            Integer(),
+            Numeric(8, 0),
+            True,
+            config.requirements.integer_subtype_comparisons,
```
sqla-tester: **mike bayer** (zzzeek) wrote:

this requirement apparently refers to just "is oracle", where we couldn't compare BigInteger to Integer.  however, now that we compare the "scale" value, this should work now?

alembic/ddl/impl.py
pbecotte: The reason I picked zip was in case was in case the lists weren't of equal length. We only compare up to the length of the shorter list, ignoring differences in "extra" arguments (that may have been added by the server defaults). If the shorter list is len 0, this loop will just be skipped.

docs/build/autogenerate.rst
pbecotte: good one!

tests/test_autogen_diffs.py
pbecotte: Unfortunately not with the other rules. Integer and Numeric come through as synonyms on Oracle, and the "metadata" column has no arguments...so we ignore the arguments on the "inspect" column.

alembic/ddl/impl.py
zzzeek: do we have any scenarios that exercise these cases?  I';m not sure our datatypes have so much positional arguments for this to matter.  I'm fine with it for now.

pbecotte:
> **mike bayer** (zzzeek) wrote:
> 
> > (1 comment)
> > pbecotte@github wrote:
> 
> alembic has been on black for many months now, actually the lines here would reformat because of the "zimports" tool I wrote first and foremost. Just turn on the pre-commit hooks and it will all be taken care of 👍 

Ah, I see- only new lines are formatted, didn't realize!

sqla-tester:
Patchset 1a6a3860881081c3b98fdbfc203f8d3af5f98926 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1561

alembic/ddl/impl.py
pbecotte: If I do

```
if inspected_params.args and meta_params.args and inspected_params.args != meta_params.args:
    return False
```

AutogenerateVariantCompareTest_oracle+cx_oracle_11_2_0_2_0::test_variant_no_issue fails comparing BigInteger to BigInteger, while test_autogen_diffs.py::CompareMetadataToInspectorTest_oracle+cx_oracle_11_2_0_2_0::test_introspected_columns_match_metadata_columns[cola10] fails comparing Boolean to Boolean. 
This is because the Oracle dialect prints those as NUMERIC(19) ... but the server adds precision to the returned value.

I added an extra test to be a bit more explicit with DECIMAL(10, 2) compared to NUMERIC(10)

sqla-tester:

next thing this needs is a file docs/build/unreleased/619.rst that describes the change.

View this in Gerrit at https://gerrit.sqlalchemy.org/1561

sqla-tester:

Code-Review+2 Workflow+1

View this in Gerrit at https://gerrit.sqlalchemy.org/1561

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1561 has been **merged**. Congratulations! :)

CaselIT:
This PR has completely removed support for `compare_against_backend`, but it's still documented and supposedly testes (the test is clearly broken.)

@zzzeek For me we can just drop it since no one noticed

CaselIT:
for reference it was originally added in dabc7f0932dd13104411469fdc5327474b2bc8e3

zzzeek:
how come tests did not fail?

CaselIT:
I haven't checked it. I'll open issue and fix this, also rewriting the tests

Issue #628, opened 2019-11-21
Migration script wrongly generated when table name is all uppercase


sergiomafra:
I'm developing an API with Flask and I'm using flask-sqlachemy to create models and flask-migrate to generate migration scripts and upgrade the database. It is an Oracle database.

The first script ran smoothly and created the tables. But the next one where I changed a constraint on a column of a tablel wasn't created right. It is trying to create again all the tables and trying to delete the same tables with lowercase letters.

My guess is when I create tables with uppercase letters, the script transforms it to lowercase at somepoint and check if it exists as a lowercased string. The tables were created fine with all uppercase letters.

Model example:

`class Blacklist(db.Model):
''' IPs from users that continuously offend the system end up here '''

__bind_key__ = None
__tablename__ = 'SCR_BLACKLIST'

ID_BLACKED = db.Column(db.Integer, primary_key=True)
IP = db.Column(db.String(15), nullable=False)
BLACKLIST_INSERTION_TYPE = db.Column(
    db.Integer,
    db.ForeignKey('SCR_BLACKLIST_INSERTION_TYPES.ID_BLACK_TIPO'),
    nullable=False)
RATIO_LIMIT = db.Column(db.Integer, nullable=False, default=0)
IP_STATUS = db.Column(db.Boolean, nullable=False)
INSERTION_DATE = db.Column(db.DateTime, nullable=False, default=tb.now())
REMOVED_DATE = db.Column(db.DateTime, nullable=False)

def __repr__(self):
    print(f'<Blacklisted IP: {self.IP}>')`

Is there a way to workaround or solve this issue?

**OS:** Ubuntu 18.04
**Database:** Oracle
**Packages:** flask-sqlalchemy, flask-migrate

zzzeek:
hi -

it is extremely unusual to have a tablename that is case **sensitive** UPPERCASE on Oracle.  were these tables created using quotes for the names ?  that is, 'CREATE TABLE "UPPERCASE_NAME" ' ? 

Otherwise, assuming these are normal, case insensitive names that you just happen to observe as UPPERCASE in your Oracle console, that is normal - specify table names in all lowercase on the SQLAlchemy side.    see https://docs.sqlalchemy.org/en/13/dialects/oracle.html#identifier-casing

PR #629, opened 2019-11-24
add a test to verify that case insensitive table names dont generate …

pbecotte:
…spurious autogen. Wrote this to check #628 - no code changes were necessary, so probably not worth actually merging this, but figured I'd offer the PR anyway :)

### DIFF
--- a/tests/test_autogen_diffs.py
+++ b/tests/test_autogen_diffs.py
@@ -15,6 +15,7 @@ from sqlalchemy import Integer
 from sqlalchemy import MetaData
 from sqlalchemy import Numeric
 from sqlalchemy import PrimaryKeyConstraint
+from sqlalchemy import Sequence
 from sqlalchemy import SmallInteger
 from sqlalchemy import String
 from sqlalchemy import Table
@@ -1187,7 +1188,7 @@ class CompareMetadataTest(ModelOne, AutogenTest, TestBase):
         )
 
 
-class PGCompareMetaData(ModelOne, AutogenTest, TestBase):
+class PGCompareMetaDataTest(ModelOne, AutogenTest, TestBase):
     __only_on__ = "postgresql"
     __backend__ = True
     schema = "test_schema"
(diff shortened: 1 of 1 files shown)


sqla-tester:
New Gerrit review created for change 97803b34788d61a63b459226b7229d40cb8326a7: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1592

sqla-tester:

sorry, the alembic tests are pretty crude as my testing-fu was not at master levels when i wrote these tests :)

pbecotte:
Interesting, in Jenkins the tests showed exactly the behavior I was trying to prove didn't happen haha. Problem is, I ran them against Oracle before submitting the PR. I need to up my Oracle-Fu ... take 6 minutes for the damn thing to start up, and it periodically gets into a state that every query takes 5 seconds and I have to delete it and start over.

zzzeek:
Im not sure things will work in Oracle (and Firebird ) if you are setting up case-sensitive UPPERCASE names on both sides.   nobody actually does this in practice, so that's good.

zzzeek:
right so...if we ask Oracle for the table "sometable", it is stored in the oracle catalog tables as SOMETABLE, so we ask, and we get back...."SOMETABLE".     If we are given case sensitive "SOMETABLE", again we ask Oracle for "SOMETABLE" and we get back..."SOMETABLE".   So there's no way for SQLAlchemy to know, when it gets back a plain list of table names, when it sees "SOMETABLE" to know if that is case sensitive "SOMETABLE" or case insensitive SOMETABLE, unless we can locate some new field in Oracle that tells us this.    this would have to be some fairly intricate change on the SQLAlchemy side that would impact how reflection works at a core level.

zzzeek:
annnndddd...this is becasue there is no "case sensitivity" in how the name is stored:

```
from sqlalchemy import create_engine

e = create_engine(
    "oracle://scott:tiger@oracle1120",
    echo='debug'
)
with e.connect() as conn:
    conn.execute('CREATE TABLE "FOO" (id integer primary key)' )

    # works, case insensitive name matches
    conn.execute("SELECT * from foo")

    # fails, because "foo" already exists
    conn.execute('CREATE TABLE foo (id integer primary key)' )

```

meaning, you cannot have a table name called "ALL_UPPERCASE" on Oracle and have it match in a case-sensitive fashion.  the test makes no sense for Oracle.

zzzeek:
by "match" i mean, reflect it back, unless there were some special client side flag we were using to make it do that when we do get_table_names(), it gets a little crazy.   just use all lower case names :)

zzzeek:
it works like this though...not sure if we have a test like this but I would think there's one in there
```

        Table(
            'MyTable',
            m,
            Column('Ident', Integer, Sequence('Id_Seq'), primary_key=True),
        )

```

those names are unambiguously case sensitive

pbecotte:
Sounds good- I think I misunderstood some of the code. Glad for the learning experience though :)

sqla-tester:

closed upstream

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1592 has been **abandoned**.  That means that at least for the moment I need to close this pull request. Sorry it didn't work out :(

PR #631, opened 2019-11-30
Add support for computed columns

### LINKED ISSUE #624: Computed column migration file generated incorrectly
SQLAlchemy version 1.3.11 (just released on November 11th, 2019) includes support for [Computed columns](https://docs.sqlalchemy.org/en/13/core/defaults.html).  From their documentation an example Computed column definition would be something like:

`Column("area", Integer, Computed("side * side"))`

After adding such a Computed column definition to my table model and trying to use alembic to autogenerate a migration file I encounter the following error:

```  File "/home/dev.internal/developer/sandbox/mbs/venv/lib/python3.7/site-packages/alembic/script/revision.py", line 321, in get_revisions
    resolved_id, branch_label = self._resolve_revision_number(id_)
  File "/home/dev.internal/developer/sandbox/mbs/venv/lib/python3.7/site-packages/alembic/script/revision.py", line 501, in _resolve_revision_number
    self._revision_map
  File "/home/dev.internal/developer/sandbox/mbs/venv/lib/python3.7/site-packages/alembic/util/langhelpers.py", line 230, in __get__
    obj.__dict__[self.__name__] = result = self.fget(obj)
  File "/home/dev.internal/developer/sandbox/mbs/venv/lib/python3.7/site-packages/alembic/script/revision.py", line 123, in _revision_map
    for revision in self._generator():
  File "/home/dev.internal/developer/sandbox/mbs/venv/lib/python3.7/site-packages/alembic/script/base.py", line 112, in _load_revisions
    script = Script._from_filename(self, vers, file_)
  File "/home/dev.internal/developer/sandbox/mbs/venv/lib/python3.7/site-packages/alembic/script/base.py", line 906, in _from_filename
    module = util.load_python_file(dir_, filename)
  File "/home/dev.internal/developer/sandbox/mbs/venv/lib/python3.7/site-packages/alembic/util/pyfiles.py", line 98, in load_python_file
    module = load_module_py(module_id, path)
  File "/home/dev.internal/developer/sandbox/mbs/venv/lib/python3.7/site-packages/alembic/util/compat.py", line 173, in load_module_py
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 724, in exec_module
  File "<frozen importlib._bootstrap_external>", line 860, in get_code
  File "<frozen importlib._bootstrap_external>", line 791, in source_to_code
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/dev.internal/developer/sandbox/mbs/migrations/versions/9396eb21e075_tip_add_user_index_generated_user_.py", line 21
    op.add_column('usertable', sa.Column('user', sa.String(length=36), server_default=Computed(<sqlalchemy.sql.elements.TextClause object at 0x7febef2d30f0>, persisted=False), nullable=True))
```  

I found that that if I hacked in a __str__() implementation to the Computed class definition in SQLAlchemy I was able to use alembic successfully however I'm not sure if that's the most elegant solution.  

I'm using alembic version 1.3.0, sqlalchemy 1.3.11 (and MySQL 8.0.17 although I don't believe that matters as far as this issue is concerned).

I really appreciate any help or suggestions!

### LINKED ISSUE #628: Migration script wrongly generated when table name is all uppercase

CaselIT:
Add alembic support for computed columns

This is still work in progress

Fix #624

### DIFF
--- a/alembic/testing/requirements.py
+++ b/alembic/testing/requirements.py
@@ -140,3 +140,14 @@ class SuiteRequirements(Requirements):
     @property
     def alter_column(self):
         return exclusions.open()
+
+    @property
+    def db_computed_columns(self):
+        # TODO: should import the requirements from sqlalchemy
+        return exclusions.skip_if(["postgresql < 12", "sqlite", "mysql < 5.7"])
+
+    @property
+    def sa_computed_column(self):
+        return exclusions.only_if(
+            exclusions.BooleanPredicate(sqla_compat.has_computed)
+        )
--- a/tests/test_autogen_computed.py
+++ b/tests/test_autogen_computed.py
@@ -9,9 +9,15 @@ from alembic.testing import exclusions
 from alembic.testing import TestBase
 from ._autogen_fixtures import AutogenFixtureTest
 
+pg_reason = (
+    "sqlalchemy on postgres detects a computed column "
+    "as a normal server default during inspection"
+)
 
-class AutogenerateComputedNoBackendTest(AutogenFixtureTest, TestBase):
-    __requires__ = ("sa_computed_column",)
+
+class AutogenerateComputedTest(AutogenFixtureTest, TestBase):
+    __requires__ = ("sa_computed_column", "db_computed_columns")
+    __backend__ = True
 
     def test_add_computed_column(self):
         m1 = MetaData()
@@ -39,10 +45,41 @@ class AutogenerateComputedNoBackendTest(AutogenFixtureTest, TestBase):
         assert c.persisted is None
         eq_(str(c.sqltext), "5")
 
+    def _test_remove_computed_column(self):
+        m1 = MetaData()
+        m2 = MetaData()
 
-class AutogenerateComputedBackendTest(AutogenFixtureTest, TestBase):
-    __requires__ = ("sa_computed_column", "db_computed_columns")
-    __backend__ = True
+        Table(
+            "user",
+            m1,
+            Column("id", Integer, primary_key=True),
+            Column("foo", Integer, sa.Computed("5")),
+        )
+
+        Table(
+            "user", m2, Column("id", Integer, primary_key=True),
+        )
+
+        diffs = self._fixture(m1, m2)
+
+        eq_(diffs[0][0], "remove_column")
+        eq_(diffs[0][2], "user")
+        c = diffs[0][3]
+        eq_(c.name, "foo")
+
+        assert c.computed is None
+        return c
+
+    @exclusions.skip_if(["postgresql"], pg_reason)
+    def test_remove_computed_column(self):
+        column = self._test_remove_computed_column()
+        assert column.server_default is None
+
+    @exclusions.only_if(["postgresql"], pg_reason)
+    def test_remove_computed_column_postgres(self):
+        column = self._test_remove_computed_column()
+        assert isinstance(column.server_default, sa.DefaultClause)
+        eq_(str(column.server_default.arg.text), "5")
 
     def _test_computed_unchanged(self, argBefore, argAfter):
         # no combination until sqlalchemy 1.3.7
@@ -77,22 +114,24 @@ class AutogenerateComputedBackendTest(AutogenFixtureTest, TestBase):
             [sa.Computed("bar*5")], [sa.Computed("bar*5")]
         )
 
+    def test_unchanged_change_expression(self):
+        self._test_computed_unchanged(
+            [sa.Computed("bar*5")], [sa.Computed("bar * 42")]
+        )
+
+    def test_unchanged_change_persisted(self):
+        self._test_computed_unchanged(
+            [sa.Computed("bar*5")], [sa.Computed("bar * 42", persisted=True)]
+        )
+
     def test_unchanged_add_computed(self):
         self._test_computed_unchanged([], [sa.Computed("bar*5")])
 
-    @exclusions.skip_if(
-        ["postgresql"],
-        "sqlalchemy on postgres detects a computed column "
-        "as a normal server default during inspection",
-    )
+    @exclusions.skip_if(["postgresql"], pg_reason)
     def test_unchanged_remove_computed(self):
         self._test_computed_unchanged([sa.Computed("bar*5")], [])
 
-    @exclusions.only_if(
-        ["postgresql"],
-        "sqlalchemy on postgres detects a computed column "
-        "as a normal server default during inspection",
-    )
+    @exclusions.only_if(["postgresql"], pg_reason)
     def test_remove_computed_postgresql(self):
         m1 = MetaData()
         m2 = MetaData()
--- a/tests/test_autogen_computed.py
+++ b/tests/test_autogen_computed.py
@@ -4,6 +4,7 @@ from sqlalchemy import Integer
 from sqlalchemy import MetaData
 from sqlalchemy import Table
 
+from alembic.testing import config
 from alembic.testing import eq_
 from alembic.testing import exclusions
 from alembic.testing import TestBase
@@ -171,4 +172,7 @@ class AutogenerateComputedTest(AutogenFixtureTest, TestBase):
 
         assert new is None
         assert isinstance(old, sa.DefaultClause)
-        eq_(str(old.arg.text), "(bar + 42)")
+        if config.db.name == "postgresql":
+            eq_(str(old.arg.text), "(bar + 42)")
+        else:  # oracle
+            eq_(str(old.arg.text), '"BAR"+42')
--- a/tests/test_autogen_computed.py
+++ b/tests/test_autogen_computed.py
@@ -5,6 +5,7 @@ from sqlalchemy import MetaData
 from sqlalchemy import Table
 
 from alembic.testing import eq_
+from alembic.testing import exclusions
 from alembic.testing import TestBase
 from ._autogen_fixtures import AutogenFixtureTest
 
@@ -38,3 +39,88 @@ class AutogenerateComputedNoBackendTest(AutogenFixtureTest, TestBase):
         assert c.persisted is None
         eq_(str(c.sqltext), "5")
 
+
+class AutogenerateComputedBackendTest(AutogenFixtureTest, TestBase):
+    __requires__ = ("sa_computed_column", "db_computed_columns")
+    __backend__ = True
+
+    def _test_computed_unchanged(self, argBefore, argAfter):
+        # no combination until sqlalchemy 1.3.7
+        m1 = MetaData()
+        m2 = MetaData()
+
+        Table(
+            "user",
+            m1,
+            Column("id", Integer, primary_key=True),
+            Column("bar", Integer),
+            Column("foo", Integer, *argBefore),
+        )
+
+        Table(
+            "user",
+            m2,
+            Column("id", Integer, primary_key=True),
+            Column("bar", Integer),
+            Column("foo", Integer, *argAfter),
+        )
+
+        diffs = self._fixture(m1, m2)
+
+        eq_(len(diffs), 0)
+
+    def test_unchanged_same_text(self):
+        self._test_computed_unchanged([sa.Computed("5")], [sa.Computed("5")])
+
+    def test_unchanged_same_text_column(self):
+        self._test_computed_unchanged(
+            [sa.Computed("bar*5")], [sa.Computed("bar*5")]
+        )
+
+    def test_unchanged_add_computed(self):
+        self._test_computed_unchanged([], [sa.Computed("bar*5")])
+
+    @exclusions.skip_if(
+        ["postgresql"],
+        "sqlalchemy on postgres detects a computed column "
+        "as a normal server default during inspection",
+    )
+    def test_unchanged_remove_computed(self):
+        self._test_computed_unchanged([sa.Computed("bar*5")], [])
+
+    @exclusions.only_if(
+        ["postgresql"],
+        "sqlalchemy on postgres detects a computed column "
+        "as a normal server default during inspection",
+    )
+    def test_remove_computed_postgresql(self):
+        m1 = MetaData()
+        m2 = MetaData()
+
+        Table(
+            "user",
+            m1,
+            Column("id", Integer, primary_key=True),
+            Column("bar", Integer),
+            Column("foo", Integer, sa.Computed("bar+42")),
+        )
+
+        Table(
+            "user",
+            m2,
+            Column("id", Integer, primary_key=True),
+            Column("bar", Integer),
+            Column("foo", Integer),
+        )
+
+        diffs = self._fixture(m1, m2)
+
+        eq_(diffs[0][0][0], "modify_default")
+        eq_(diffs[0][0][2], "user")
+        eq_(diffs[0][0][3], "foo")
+        old = diffs[0][0][-2]
+        new = diffs[0][0][-1]
+
+        assert new is None
+        assert isinstance(old, sa.DefaultClause)
+        eq_(str(old.arg.text), "(bar + 42)")
--- a/tests/test_autogen_computed.py
+++ b/tests/test_autogen_computed.py
@@ -1,0 +1,40 @@
+import sqlalchemy as sa
+from sqlalchemy import Column
+from sqlalchemy import Integer
+from sqlalchemy import MetaData
+from sqlalchemy import Table
+
+from alembic.testing import eq_
+from alembic.testing import TestBase
+from ._autogen_fixtures import AutogenFixtureTest
+
+
+class AutogenerateComputedNoBackendTest(AutogenFixtureTest, TestBase):
+    __requires__ = ("sa_computed_column",)
+
+    def test_add_computed_column(self):
+        m1 = MetaData()
+        m2 = MetaData()
+
+        Table(
+            "user", m1, Column("id", Integer, primary_key=True),
+        )
+
+        Table(
+            "user",
+            m2,
+            Column("id", Integer, primary_key=True),
+            Column("foo", Integer, sa.Computed("5")),
+        )
+
+        diffs = self._fixture(m1, m2)
+
+        eq_(diffs[0][0], "add_column")
+        eq_(diffs[0][2], "user")
+        eq_(diffs[0][3].name, "foo")
+        c = diffs[0][3].computed
+
+        assert isinstance(c, sa.Computed)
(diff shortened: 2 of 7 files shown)


CaselIT:
I've made a typo in the branch name 😞

alembic/testing/requirements.py
```diff
@@ -140,3 +140,14 @@ def comments_api(self):
     @property
     def alter_column(self):
         return exclusions.open()
+
+    @property
+    def db_computed_columns(self):
+        # TODO: should import the requirements from sqlalchemy
+        return exclusions.skip_if(["postgresql < 12", "sqlite", "mysql < 5.7"])
```
CaselIT: @zzzeek here I should ideally return the value of `sqlalchemy.testing.requirements.SuiteRequirements.computed_columns` but currently it defines a `closed` condition. 
The real condition is set in `test.requirements.DefaultRequirements.computed_columns` which I cannot import here

Maybe the condition to indicate if a backend supports computed columns should be somewhere in the `sqlalchemy.testing` library so that it can be imported?

CaselIT:
I think I'm stuck.
I'm not sure if it is possible to implement something that works reliably without inspect support for computed columns. I think that best outcome for now is to just ignore the computed column so it will not mess with the rest.

The current implementation in this pr works for (in pg and mysql):

- adding a table with a computed column
- adding a computed column to an existing table

The problems it has are (that I found):

- after adding a computed column the next migration without any change to that column will:
  - in mysql:
    ```py
    def upgrade():
        op.alter_column("square", "side", existing_type=mysql.FLOAT(), server_default="0", existing_nullable=True)
    
    def downgrade():
        op.alter_column("square", "side", existing_type=mysql.FLOAT(), server_default=sa.text("'0'"), existing_nullable=True)
    ```
  - in pg: 
    ```py
    def upgrade():
        op.alter_column(
            "square", "area", existing_type=postgresql.DOUBLE_PRECISION(precision=53), server_default=sa.Computed("side * side",), existing_nullable=True
        )    

    def downgrade():
        # This will raise error in pg when executed
        op.alter_column(
            "square", "area", existing_type=postgresql.DOUBLE_PRECISION(precision=53), server_default=sa.text("(side * side)"), existing_nullable=True
        )
    ```

- dropping a computed column will renders the wrong column in downgrade
  - in mysql:
  ```py
  def downgrade():
    op.add_column("square", sa.Column("area", mysql.FLOAT(), nullable=True))
  ```
  - in pg
  ```py
  def downgrade():
    # This will raise error in pg when executed
    op.add_column(
        "square",
        sa.Column("area", postgresql.DOUBLE_PRECISION(precision=53), server_default=sa.text("(side * side)"), autoincrement=False, nullable=True),
    )
  ```

In this branch https://github.com/CaselIT/alembic/tree/support_compute_changes I've tried to make the server default work with postgress, but it will break in mysql since its behaviour is different. I'm down think it is the right approach

CaselIT:
found another issue

- changing a computed will be detected sometimes but it not work
  - in mysql:
  ```py
  def upgrade():
      op.alter_column("square", "side", existing_type=mysql.FLOAT(), server_default="0", existing_nullable=True)

  def downgrade():
      op.alter_column("square", "side", existing_type=mysql.FLOAT(), server_default=sa.text("'0'"), existing_nullable=True)
  ```
  - in pg
  ```py
  def upgrade():
    # This will raise error in pg when executed because there is not support for altering the computed
    op.alter_column(
        "square", "area", existing_type=postgresql.DOUBLE_PRECISION(precision=53), server_default=sa.Computed("side + side",), existing_nullable=True
    )

  def downgrade():
    # This will raise error in pg when executed
    op.alter_column(
        "square", "area", existing_type=postgresql.DOUBLE_PRECISION(precision=53), server_default=sa.text("(side * side)"), existing_nullable=True
    )
  ```

zzzeek:
all the cases that would generate an "alter" should be ignored for now.   when the metadata column has a computed object on it, compare_server_default needs to return "no change", so that there is no false positive.

CaselIT:
the downgrade error on pg will not be solved though

zzzeek:
> the downgrade error on pg will not be solved though

OK, in that the server default is an intrinsic part of the column makes it a little more inconvenient than a separate constraint, however, if/when SQLAlchemy implements reflection for generated columns this part in Alembic will just start working, so it's not making things worse to implement rendering for starters.

in the world of database migrations, there's a few operations that are common, frequent, and important, and others that are less common, infrequent, not as important.

the important cases are:

1. create table
2. create index
3. alter table add column
4. alter table add constraint
5. adding ENUM members

most of which are detected right now and rendered correctly with the exception of a CHECK constraint being added and kind of critically the ENUM one, which is a long running quagmire kind of problem.

the non important cases are:

1. everything else - changing column type, changing column default, adding/removing column default, changing cols inside an index or constraint, changing index expressions, changing nullable, changing table name, changing column name, etc.    this is because these operations are not additive and break backwards compatibility so they are usually more awkward as the corresponding applications have to change as well, therefore they are less frequent and when they are undertaken, it's typical people have to edit their migration files anyway.

CaselIT:
Ok @zzzeek, I trust your judgment on this.

I've updated the implementation to make alembic ignore changes when adding or removing computed to an existing column. It also ignores if the computed expression changes or the persistence changes.

It should be ready for review

There is still the error in postgresql when removing a computed column or computed from column, but I'm not sure currently there is a way to avoid it. I think a note in the documentation will have to suffice for now
I've made special tests for it, to test the current (erroneous) behaviour.

I've only tested on postgres and mysql, so I guess there may be some other surprises with other backends

To recap, in postgres the current implementation renders a erroneous downgrade (that will error if executed) when:
- removing a computed column
- removing computed from a column (without removing the column)

In both cases the problem is that the server default will be set to the expression that was in the computed class

sqla-tester:
New Gerrit review created for change 7726d0957681634f9c772355463acb530d520462: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1600

zzzeek:
OK ill start moving it in the queue.   we should absolutely try to implement reflection for GENERATED in SQLAlchemy also, it's just, implementing what we have so far in Alembic is not going to make things worse so we might as well.

CaselIT:
I must have left a type somewhere. Also it seems that oracle behaves like pg

CaselIT:
I think the new revision should pass in py2.7 and oracle

zzzeek:
is this one ready for review?

CaselIT:
I think so

sqla-tester:
Patchset 16c760cb3fe6d55840d025bacabcb4188df0bc59 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1600

CaselIT:
oh nice, it seems I'm able to reask to sqla-tester to run the tests once it has be assigned to a pull request.

sqla-tester:
Patchset b1884461addb0c39a7d86ddc8e1139ab8ba1fa78 added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1600

sqla-tester:
Patchset 9c45651295626edf5f172ec827a5ced1440818cf added to existing Gerrit review https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1600

CaselIT:
all the tests should pass now. the previous build had only pep8 failures

sqla-tester:

hey....I think there is no DDL to actually alter a column that does not have a "generated" to have one, and vice versa.  I think you can only drop or add columns.    Which would render the whole "detection" issue sort of moot.

Looking at https://www.postgresql.org/docs/12/sql-altertable.html and trying everyhitng I can on the PG commandline, I can't add "GENERATED AS" to an existing column.    see https://dba.stackexchange.com/a/250869.

I'm working on this now and I think what I will do is that if changes to "computed" are detected with that server default thing, ignore it always, and if someone actually puts "Computed" in an op.alter_column() it has to raise.

CaselIT:
That makes sense.
If I remember correctly the problem is that the generated column gets detected as a default in pg and oracle, so it's harder to discriminate it from an actual default value.

I don't know regarding the ddl if other backend support altering an existing column. Regarding pg they may be adding it in a later version, since iirc this version only had initial support for grnerated columns.

tests/test_autogen_computed.py
```diff
@@ -0,0 +1,178 @@
...
+
+        diffs = self._fixture(m1, m2)
+
+        eq_(diffs[0][0], "remove_column")
+        eq_(diffs[0][2], "user")
+        c = diffs[0][3]
+        eq_(c.name, "foo")
+
+        assert c.computed is None
+        return c
```
sqla-tester: **Federico Caselli** (CaselIT) wrote:

nice we now have combinations also here

CaselIT:
Maybe I could open an issue on sqlalchemy to track the inspection of computed columns

sqla-tester:

Thanks for the review Mike

zzzeek:
> Maybe I could open an issue on sqlalchemy to track the inspection of computed columns

sure, the days of me keeping the number of open issues below 150 are long over .... :)

CaselIT:
> sure, the days of me keeping the number of open issues below 150 are long over .... :)

I can wait :P

sqla-tester:

just a quick and easy drop in of Computed, no biggie....yeah right

sqla-tester:

> Patch Set 8:
> 
> just a quick and easy drop in of Computed, no biggie....yeah right

yes, this got quite a lot of bigger that the 4 lines required to render the computed string in add column :\

CaselIT:
maybe we should just wait until sqlalchemy support inspecting the computed columns. 
Or at the very least do not report them as server_default in some backends?

zzzeek:
> maybe we should just wait until sqlalchemy support inspecting the computed columns.
> Or at the very least do not report them as server_default in some backends?

i don't think it's really going to be that much of a problem.    putting in the render for Computed in a model is still essential whether or not detection works.

zzzeek:
also a bunch of complexity here is because the combinations thing has proven to be so useful and I'm not ready to drop SQLA 1.2 support in alembic yet.

CaselIT:
btw my suggestion was mainly regarding the workaround for server_default. Once it's not needed the code still needs to be there so support the sqlalchemy version that has computed but not the correct detection

CaselIT:
Avoiding returning the default values seems very easy.

this is a patch for pg
```diff
diff --git a/lib/sqlalchemy/dialects/postgresql/base.py b/lib/sqlalchemy/dialects/postgresql/base.py
index 6d97033d0..1b0b2b46a 100644
--- a/lib/sqlalchemy/dialects/postgresql/base.py
+++ b/lib/sqlalchemy/dialects/postgresql/base.py
@@ -2827,7 +2827,14 @@ class PGDialect(default.DefaultDialect):
         table_oid = self.get_table_oid(
             connection, table_name, schema, info_cache=kw.get("info_cache")
         )
-        SQL_COLS = """
+
+        generated = (
+            "a.attgenerated as generated"
+            if self.server_version_info >= (12,)
+            else "'' as generated"
+        )
+        SQL_COLS = (
+            """
             SELECT a.attname,
               pg_catalog.format_type(a.atttypid, a.atttypmod),
               (SELECT pg_catalog.pg_get_expr(d.adbin, d.adrelid)
@@ -2836,7 +2843,8 @@ class PGDialect(default.DefaultDialect):
                AND a.atthasdef)
               AS DEFAULT,
               a.attnotnull, a.attnum, a.attrelid as table_oid,
-              pgd.description as comment
+              pgd.description as comment,
+              %s
             FROM pg_catalog.pg_attribute a
             LEFT JOIN pg_catalog.pg_description pgd ON (
                 pgd.objoid = a.attrelid AND pgd.objsubid = a.attnum)
@@ -2844,6 +2852,8 @@ class PGDialect(default.DefaultDialect):
             AND a.attnum > 0 AND NOT a.attisdropped
             ORDER BY a.attnum
         """
+            % generated
+        )
         s = (
             sql.text(SQL_COLS)
             .bindparams(sql.bindparam("table_oid", type_=sqltypes.Integer))
@@ -2876,6 +2886,7 @@ class PGDialect(default.DefaultDialect):
             attnum,
             table_oid,
             comment,
+            generated
         ) in rows:
             column_info = self._get_column_info(
                 name,
@@ -2886,6 +2897,7 @@ class PGDialect(default.DefaultDialect):
                 enums,
                 schema,
                 comment,
+                generated,
             )
             columns.append(column_info)
         return columns
@@ -2900,6 +2912,7 @@ class PGDialect(default.DefaultDialect):
         enums,
         schema,
         comment,
+        generated,
     ):
         def _handle_array_type(attype):
             return (
@@ -3010,6 +3023,11 @@ class PGDialect(default.DefaultDialect):
                 "Did not recognize type '%s' of column '%s'" % (attype, name)
             )
             coltype = sqltypes.NULLTYPE
+
+        # remove default from generated columns
+        if generated != '':
+            default = None
+
         # adjust the default value
         autoincrement = False
         if default is not None:
```

CaselIT:
oracle (seems to require `all_tab_cols` instead of `all_tab_columns`)

```diff
diff --git a/lib/sqlalchemy/dialects/oracle/base.py b/lib/sqlalchemy/dialects/oracle/base.py
index fe3d586a4..d41f0ed34 100644
--- a/lib/sqlalchemy/dialects/oracle/base.py
+++ b/lib/sqlalchemy/dialects/oracle/base.py
@@ -1658,8 +1658,8 @@ class OracleDialect(default.DefaultDialect):
         text = """
             SELECT col.column_name, col.data_type, col.%(char_length_col)s,
               col.data_precision, col.data_scale, col.nullable,
-              col.data_default, com.comments\
-            FROM all_tab_columns%(dblink)s col
+              col.data_default, com.comments, col.virtual_column\
+            FROM all_tab_cols%(dblink)s col
             LEFT JOIN all_col_comments%(dblink)s com
             ON col.table_name = com.table_name
             AND col.column_name = com.column_name
@@ -1684,6 +1684,7 @@ class OracleDialect(default.DefaultDialect):
             nullable = row[5] == "Y"
             default = row[6]
             comment = row[7]
+            genetared = row[8]
 
             if coltype == "NUMBER":
                 if precision is None and scale == 0:
@@ -1708,6 +1709,9 @@ class OracleDialect(default.DefaultDialect):
                     )
                     coltype = sqltypes.NULLTYPE
 
+            if genetared == 'YES':
+                default = None
+
             cdict = {
                 "name": colname,
                 "type": coltype,
```

CaselIT:
> hey....I think there is no DDL to actually alter a column that does not have a "generated" to have one, and vice versa. I think you can only drop or add columns. Which would render the whole "detection" issue sort of moot.

I've thought a bit about it and it may make sense that the expected way to modify a generated colum is to drop it and to regenerate it. Since it is generated it should usually be pretty safe to do*.
I'm not sure if alembic should do something like this though (I mean drop then re-add the column with the new generation)

* this in the backends that do not let the generated column expression contain other generated columns. From memory pg does not allow this, but mysql does. So the safeness of this operation may depend on the backend.

CaselIT:
Maybe this logic (deop+readd) could be under a flag/configuration options (disabled by default)

But I'm not sure if this complexity is worth it

zzzeek:
> Maybe this logic (drop+readd) could be under a flag/configuration options (disabled by default)
> 
> But I'm not sure if this complexity is worth it

not right now, I dont think this is going to be a widely used feature and autogenerate is not intended to be fully automatic.

zzzeek:
it would help if I added one of the .py files to the gerrit

CaselIT:
Maybe creating an external package for the testing stuff could be useful

zzzeek:
the "combinations" thing was a huge change to the whole way the testing package works, usually things are not this out of sync.    I worry that having to put new releases of a separate testing package every time I need to fix something is going to make things more complicated, not to mention if the testing package itself relies upon SQLAlchemy features I'd have to be maintaining several branches of the testing package at the same time.

CaselIT:
yes I agree that it can create some friction. maybe just the general framework that does not depend specifically on sqlalchemy could go there, like the catch warnings-errors, the assertions (eq_, is_, etc) and combinations.

but I guess the current situation is temporary since I guess alembic will not keep supporting version 1.1 and 1.2 forever

zzzeek:
since the testing framework is also consumed by third party dialects and not just Alembic, having it be a separate package, probably one that has lots of logic and fallbacks to tell which SQLAlchemy version it runs against, is probably a good idea.   at the moment though I don't think I can handle it.

zzzeek:
that is, rather than having lots of branches of the "testing framework" just have all the different SQLAlchemy assumptions present in one big release.

CaselIT:
Let me know if there is anything else I can do for this pr

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1600 has been **merged**. Congratulations! :)

PR #639, opened 2019-12-22
Fix tests on windows

### LINKED ISSUE #632: Tests do not pass on windows
Under windows some tests of alembic fail.

The main reason for failures are 
- a Permission error `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'scratch\\scripts\\foo.db'`, when setting up some tests with sqlite
- slash/backslash difference ` call('foobar\\versions') != call('foobar/versions')`
- newline with `\n` vs `\r\n`

This is mainly to let you know, I'm not sure if it feasible to ensure that the tests work also on windows, since I guess it's hard to setup a ci environment with windows

if this is the expect behavior it can be closed

CaselIT:
Update the tests to make them run on windows
### Description

The main changes are:
- Correctly manage the paths on windows vs *nix like system.
- Manage the new lines \r\n vs \n
- Windows cannot remove open files, namely the sqlite db, so it keeps track of the connections and closes them before removing the files.

I've tested with sqlite, postgres and mysql and all pass (on windows 10)

This fixes #632 
### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/testing/env.py
+++ b/alembic/testing/env.py
@@ -13,7 +13,6 @@ from ..script import ScriptDirectory
 from ..util.compat import get_current_bytecode_suffixes
 from ..util.compat import has_pep3147
 from ..util.compat import u
-from ..util.compat import win
 
 
 def _get_staging_directory():
@@ -60,6 +59,13 @@ def staging_env(create=True, template="generic", sourceless=False):
     return sc
 
 
+def clear_staging_env():
+    from sqlalchemy.testing import engines
+
+    engines.testing_reaper.close_all()
+    shutil.rmtree(_get_staging_directory(), True)
+
+
 def script_file_fixture(txt):
     dir_ = os.path.join(_get_staging_directory(), "scripts")
     path = os.path.join(dir_, "script.py.mako")
@@ -87,53 +93,10 @@ config = context.config
         f.write(txt)
 
 
-if win:
-    # windows cannot remove an open file. An sqlite db file stays open
-    # until any connection to it is open, so we need to keep track of the
-    # engine and of all the connection to close them when clearing the
-    # the staging env
-    def clear_staging_env():
-        for entry in _engine_registry.values():
-            _clear_engine(entry)
-        _engine_registry.clear()
-        shutil.rmtree(_get_staging_directory(), True)
-
-    _engine_registry = {}
-
-    def _clear_engine(entry):
-        entry["engine"].dispose()
-        for conn in entry["connections"]:
-            conn.close()
-
-    def _sqlite_file_db(tempname="foo.db"):
-        dir_ = os.path.join(_get_staging_directory(), "scripts")
-        url = "sqlite:///%s/%s" % (dir_, tempname)
-        eng = engines.testing_engine(url=url)
-
-        if url in _engine_registry:
-            _clear_engine(_engine_registry[url])
-        entry = {"engine": eng, "connections": []}
-        _engine_registry[url] = entry
-        _orig_connect = eng.connect
-
-        def connect(*args, **kwargs):
-            conn = _orig_connect(*args, **kwargs)
-            entry["connections"].append(conn)
-            return conn
-
-        eng.connect = connect
-        return eng
-
-
-else:
-
-    def clear_staging_env():
-        shutil.rmtree(_get_staging_directory(), True)
-
-    def _sqlite_file_db(tempname="foo.db"):
-        dir_ = os.path.join(_get_staging_directory(), "scripts")
-        url = "sqlite:///%s/%s" % (dir_, tempname)
-        return engines.testing_engine(url=url)
+def _sqlite_file_db(tempname="foo.db"):
+    dir_ = os.path.join(_get_staging_directory(), "scripts")
+    url = "sqlite:///%s/%s" % (dir_, tempname)
+    return engines.testing_engine(url=url)
 
 
 def _sqlite_testing_config(sourceless=False):
--- a/alembic/testing/env.py
+++ b/alembic/testing/env.py
@@ -13,6 +13,7 @@ from ..script import ScriptDirectory
 from ..util.compat import get_current_bytecode_suffixes
 from ..util.compat import has_pep3147
 from ..util.compat import u
+from ..util.compat import win
 
 
 def _get_staging_directory():
@@ -59,10 +60,6 @@ def staging_env(create=True, template="generic", sourceless=False):
     return sc
 
 
-def clear_staging_env():
-    shutil.rmtree(_get_staging_directory(), True)
-
-
 def script_file_fixture(txt):
     dir_ = os.path.join(_get_staging_directory(), "scripts")
     path = os.path.join(dir_, "script.py.mako")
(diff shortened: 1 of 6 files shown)


alembic/testing/env.py
```diff
@@ -90,10 +87,53 @@ def env_file_fixture(txt):
...
+        _engine_registry[url] = entry
+        _orig_connect = eng.connect
+
+        def connect(*args, **kwargs):
+            conn = _orig_connect(*args, **kwargs)
+            entry["connections"].append(conn)
+            return conn
+
+        eng.connect = connect
+        return eng
```
CaselIT: I'm not sure if there is a more elegant way to get all the connections. This version seems to work but it less than elegant

zzzeek: oh.   right, so yeah, there's no need to monkeypatch, there's an event system that intercepts these.    within the testing suite, this should already be happening, that is, if you call:

    from sqlalchemy.testing import engines
    engines.testing_reaper.close_all()

this should be getting called after every test suite teardown.     it's getting called right here:

https://github.com/sqlalchemy/alembic/blob/master/alembic/testing/plugin/plugin_base.py#L28

so basically none of the above should be needed.

zzzeek: by which I mean, if we are getting a stack trace there and it is failing to remove the files, some test or fixture is not correctly using the testing.engines package to track that engine.

CaselIT: The stack trace is on "foo.db". Cannot remove open file.

The stack trace happens on basically every test that uses the db when using sqlite

CaselIT: One of the classes that has the error is this one `tests/test_command.py::StampMultipleHeadsTest`

It calls only `clear_staging_env` in its tear down

> this should be getting called after every test suite teardown. it's getting called right here: 

this seems to only work once at the end of the class, not after every test.
So it seems that `clear_staging_env` runs after every test, but the engine clear runs only at the end of the class.

I'll use `engines.testing_reaper.close_all()` in `clear_staging_env`

CaselIT:
I've updated the patch to use `engines.testing_reaper.close_all()`

alembic/testing/util.py
```diff
@@ -78,3 +80,11 @@ def run_ddl(self):
         return fixture_functions.fixture(scope=ddl)(run_ddl)
 
     return decorate
+
+
+def path(linuxPath):
```
zzzeek: oops no camels allowed :)

flake8 will probably hit this

alembic/testing/env.py
```diff
@@ -60,6 +60,9 @@ def staging_env(create=True, template="generic", sourceless=False):
 
 
 def clear_staging_env():
+    from sqlalchemy.testing import engines
+
+    engines.testing_reaper.close_all()
```
zzzeek: greeeaaat!!!!!

alembic/testing/util.py
```diff
@@ -78,3 +80,11 @@ def run_ddl(self):
...
+    """Takes a linux path with / and transforms it to the correct os path.
+    This is mainly to support windows \\"""
+    if win:
+        return join(*linuxPath.split("/"))
```
zzzeek: doesnt.....os.path.normpath do this?

tests/test_command.py
```diff
@@ -95,14 +96,17 @@ def _eq_cmd_output(self, buf, expected, env_token=False, currents=()):
...
+            entry = rev.log_entry
+            if os.linesep != '\n':
+                entry = entry.replace('\n', os.linesep)
+            assert_lines.append(entry)
 
         if env_token:
             assert_lines.insert(0, "environment included OK")
 
         eq_(
             buf.getvalue().decode("ascii", "replace").strip(),
```
zzzeek: so on this side we are taking "raw" content and converting to something agnostic, so let's make it:

    buf.getvalue().decode("ascii", "replace").replace(os.linesep, "\n").strip()    

and that's it, then we don't need a conditonal above or anything

tests/test_command.py
```diff
@@ -1113,7 +1117,7 @@ def access_(path, mode):
             command.init(self.cfg, directory="foobar")
             eq_(
                 makedirs.mock_calls,
-                [mock.call("foobar"), mock.call("foobar/versions")],
+                [mock.call("foobar"), mock.call(path("foobar/versions"))],
```
zzzeek: can we call "path" something more descriptive like "ospath("/foo/bar")"

tests/test_script_production.py
```diff
@@ -1248,7 +1248,9 @@ def test_env_emits_warning(self):
                     )
                 )
```
zzzeek: would we...do the replace("\\", ..) on the path part here and that's it?  I'm not understanding this escaping should we just make it an r"" string?

alembic/testing/util.py
CaselIT: I usually code in camels so sometimes I use them out of habit

alembic/testing/util.py
CaselIT: Indeed, I remembered there was something that did this, but I did not remember what it was.

tests/test_command.py
CaselIT: I preferred to change the expected value, but in this case I guess it's best change the actual value

tests/test_script_production.py
CaselIT: the problem is that `assertions.expect_warnings(mgs)` transforms `mgs` (typo.. 😞) to a regexp, so that causes problem.

So we have something like this

```py
msg = '''someting something
C:\\foo\\bar\\path
foo bar
'''
# in expect_warnings
re.compile(msg)
```

this does not work, since we an error like `error: bad escape \p at position 29 (line 2, column 11)`

CaselIT: I have not checked if we can disable the regexp parse in `expect_warnings`. Maybe that's the best solution

zzzeek: oh it's a reg.   OK...maybe add an expect_warnings_plain() function then.   this test is manufacturing the assertions from the data which is kind of backwards.

CaselIT:
I think I've updated all review comments

tests/test_script_production.py
```diff
@@ -1248,7 +1248,8 @@ def test_env_emits_warning(self):
...
             )
-        ):
+        )
+        with assertions.expect_warnings(msg, regex=False):
```
zzzeek: oh looky that I already had it, neat

alembic/testing/util.py
```diff
@@ -78,3 +80,11 @@ def run_ddl(self):
         return fixture_functions.fixture(scope=ddl)(run_ddl)
 
     return decorate
+
+
+def ospath(linux_path):
+    """Takes a linux path with / and transforms it to the correct os path.
+    This is mainly to support windows \\"""
+    if win:
+        return normpath(linux_path)
```
zzzeek: could we just be calling os.path.normpath in all cases, directly?    e.g. we don't need the conditional and then we don't really need "ospath".  just let normpath do it (or not).

CaselIT: I'll look if it can be used. If had some problem with the leadit / on linux. But I can update the test to suppt it

CaselIT:
replaced ospath with os.path.normpath

sqla-tester:
New Gerrit review created for change c6b7794510b5179f6ea574019df79a040e3c5f42: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1627

sqla-tester:

Code-Review+2 Workflow+1

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1627 has been **merged**. Congratulations! :)

PR #646, opened 2020-01-23
Partial ordering

ms32035:
### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/tests/test_batch.py
+++ b/tests/test_batch.py
@@ -41,7 +41,7 @@ class BatchApplyTest(TestBase):
     def setUp(self):
         self.op = Operations(mock.Mock(opts={}))
 
-    def _simple_fixture(self, table_args=(), table_kwargs={}):
+    def _simple_fixture(self, table_args=(), table_kwargs={}, **kw):
         m = MetaData()
         t = Table(
             "tname",
@@ -50,7 +50,7 @@ class BatchApplyTest(TestBase):
             Column("x", String(10)),
             Column("y", Integer),
         )
-        return ApplyBatchImpl(t, table_args, table_kwargs, False)
+        return ApplyBatchImpl(t, table_args, table_kwargs, False, **kw)
 
     def _uq_fixture(self, table_args=(), table_kwargs={}):
         m = MetaData()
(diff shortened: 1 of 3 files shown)


zzzeek:
heya.  sorry, I'm really out of it.  lets put this in CI and see what happens.   feel free to ping me if I fall asleep again.

sqla-tester:
New Gerrit review created for change 29392b796dd995fabdabe50e8725385dbe1b4883: https://gerrit.sqlalchemy.org/#/c/sqlalchemy/alembic/+/1698

tests/test_batch.py
```diff
@@ -466,6 +466,21 @@ def test_add_col(self):
...
+    def test_add_col_partial_ordering(self):
+        impl = self._simple_fixture()
+        col = Column("g", Integer)
+        # operations.add_column produces a table
+        t = self.op.schema_obj.table("tname", col)  # noqa
+        impl.add_column("tname", col, partial_ordering=("x", "g", "y"))
+        new_table = self._assert_impl(impl, colnames=["id", "x", "g", "y"])
+        eq_(new_table.c.g.name, "g")
+
+    # TODO: doesn't work, don't know why
```
sqla-tester: **mike bayer** (zzzeek) wrote:

so...this passes now and I can take that comment out

View this in Gerrit at https://gerrit.sqlalchemy.org/1698

ms32035:
> **mike bayer** (zzzeek) wrote:
> 
> (1 comment)
> 
> so I think partial_ordering is handy but I bet what you really want for the add_column() use case is a simpler "insert_after" / "insert_before" kind of thing, because partial_ordering doesn't really do that without putting out the whole list of columns, so let me see if that can be done.
> 
> View this in Gerrit at https://gerrit.sqlalchemy.org/1698

well, I guess the difference is that `insert_after/insert_before` requires just one other column assuming the current order is correct, whereas `partial_ordering` needs all the columns, but gives you more power. In my experience (primarily ETL), in most of the cases the former is sufficient, but I've also seen cases when the latter was needed.

zzzeek:
> > **mike bayer** (zzzeek) wrote:
> > (1 comment)
> > so I think partial_ordering is handy but I bet what you really want for the add_column() use case is a simpler "insert_after" / "insert_before" kind of thing, because partial_ordering doesn't really do that without putting out the whole list of columns, so let me see if that can be done.
> > View this in Gerrit at https://gerrit.sqlalchemy.org/1698
> 
> well, I guess the difference is that `insert_after/insert_before` requires just one other column assuming the current order is correct, whereas `partial_ordering` needs all the columns, but gives you more power. In my experience (primarily ETL), in most of the cases the former is sufficient, but I've also seen cases when the latter was needed.

I'm allowing both parameters, but if you want to do total reordering you do it at the batch_alter_table() call.  For the individual add_columns(), my notion that someone could just add a short ordering is too open-ended and doesn't actually work for the most common use case of adding a single column to a specific spot.   It also mirrors your original request to make it behave like MySQL's FIRST / AFTER thing.

zzzeek:
note the latest version, when you give it insert_before or insert_after, there's a whole dump of new logic in there to figure out how to order this new column with the other ones since it has to be specific the way I did this w/ the topological thing.    it's a little overbuilt to support the more open-ended use case.

ms32035:
Had a quick peek at gerrit and it looks amazing. Never seen before any db migration tool to be able to do this in an automated way. Will try out next week. Great stuff. Thanks @zzzeek

zzzeek:
this will close automatically when we're all merged

sqla-tester:

Code-Review+2 Workflow+1

View this in Gerrit at https://gerrit.sqlalchemy.org/1698

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/1698 has been **merged**. Congratulations! :)

Issue #661, opened 2020-02-21
MySQL dialect types generating spurious revisions in 1.4


5uper5hoot:
Auto-generating a revision in 1.3.3 does not pick up any changes, however after upgrading to 1.4, every MySQL dialect type column generates a revision such as this one:
```
    op.alter_column('comp_details', 'market_type_id',
               existing_type=mysql.INTEGER(display_width=10, unsigned=True),
               type_=mysql.INTEGER(unsigned=True),
               existing_nullable=True)
```
Here are the column definitions for that one:
```
    market_type_id = Column(INTEGER(unsigned=True), nullable=True)
```
I only use the mysql specific integer types, so am unsure if this manifests with other mysql dialect types but would be happy to dig deeper if it would be helpful.

zzzeek:
hey there -

this is sort of the kind of bug I figured we'd have with the new system.  the logic is supposed to not generate4 a migration if one side has an option that the other doesnt, because they can be implicit on one side or the other.  guess that's not working here but we'll add this case and it should hopefully fix this whole class of issues.

zzzeek:
cc @pbecotte , I'll take a look today

pbecotte:
Yeah, I don't think there's any tests using unsigned specifically, so dunno without looking why. I will also have time today.

pbecotte:
So- this is definitely a bug. See my PR for more details. However, it left me even more confused trying to duplicate it. I set up test cases-

```
        (
            mysql.INTEGER(unsigned=True, display_width=10),
            mysql.INTEGER(unsigned=True, display_width=10),
            False,
        ),
        (mysql.INTEGER(unsigned=True), mysql.INTEGER(unsigned=True), False),
        (
            mysql.INTEGER(unsigned=True, display_width=10),
            mysql.INTEGER(unsigned=True),
            False,
        ),
        (
            mysql.INTEGER(unsigned=True),
            mysql.INTEGER(unsigned=True, display_width=10),
            False,
        ),
```

This did show the bug. My problem is that it showed the bug on case 1 and 3 while 2 and 4 passed. This is exactly the opposite of what I expected with the bug report. Turns out no matter what I do the inspected columns were never coming back with a display_width setting even while setting it directly...while this bug report suggests that it was getting that setting even if it wasn't set.

What is your setup? sqlalchemy, mysql, and driver versions, I mean? Its possible that the testing isn't working the way I think it does :/

5uper5hoot:
[...]

PR #662, opened 2020-02-24
original comparison regex generates false positives in situations whe…

pbecotte:
…re parameters inside parentheses can mix with those outside of it.

See #661 

The previous comparison just split the string on '(,)' breaking out stuff like INTEGER(10, blah=true) into integer, 10, blah=true. Problem was with strings like INTEGER(10) UNSIGNED - it was coming out integer, 10, unsigned ... while INTEGER UNSIGNED comes out as 'integer unsigned'. 

This takes the approach of merging anything before and after the parentheses into a single term. This passes the current tests- but may not be right. It may be more correct here to split on whitespace as well as the parentheses and commas so that "INTEGER UNSIGNED" comes out as integer, unsigned so that it does not generate a diff against integer, 10, unsigned. However, the naive implementation I just described wouldn't work either since `10 != unsigned`.

My feeling is this is a quick fix to the noticed bug, and I should go through some database documentation to get an idea of all the combinations of arguments that will actually show up in real column definitions.

### DIFF
--- a/alembic/ddl/impl.py
+++ b/alembic/ddl/impl.py
@@ -330,11 +330,15 @@ class DefaultImpl(with_metaclass(ImplMeta)):
 
     def _tokenize_column_type(self, column):
         definition = self.dialect.type_compiler.process(column.type).lower()
-        # py27 - py36 - col_type, *param_terms = re.findall...
-        matches = re.findall("[^(),]+", definition)
-        col_type, param_terms = matches[0], matches[1:]
+        matches = re.search(
+            r"^(?P<col>[^()]*)(?:\((?P<params>.*?)\))?(?P<ext>[^()]*)?$",
+            definition,
+        ).groupdict(default="")
+        col_type = matches["col"]
+        if matches["ext"]:
+            col_type = col_type.strip() + " " + matches["ext"].strip()
         params = Params(col_type, [], {})
-        for term in param_terms:
+        for term in matches["params"].split(",") if matches["params"] else []:
             if "=" in term:
                 key, val = term.split("=")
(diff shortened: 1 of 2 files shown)


zzzeek:
oh damn.   sorry.  wow OK that's my work for today fixing that

sqla-tester:
New Gerrit review created for change 28f181247ac475069eb8cd3669331e689fc78792: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/1739

sqla-tester:
Patchset 28f181247ac475069eb8cd3669331e689fc78792 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/1739

zzzeek:
I love the smell of running patch inside of site-packages in the morning

sqla-tester:

recheck

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/1739

sqla-tester:

Code-Review+2 Workflow+1

IMO this makes things better, and the existing tests continue to pass including tricky ones like TIMESTAMP WITH TIMEZONE.   I tweaked it a bit but the approach is good enough for now until we get more reports from the outside.  MySQL's format is likely the weirdest we're going to see.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/1739

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/1739 has been **merged**. Congratulations! :)

Issue #789, opened 2021-02-01
downgrade with depedencies leaves revision table in broken state


hcpatel:
**Describe your question**
My use case is 2 branches with their own base to start with, where subsequent revisions from BranchB will depend on BranchA. Each new revision on BranchB I always specify a 'depends on' BranchA (usually --depend_on=branchA@head). But I an finding an issue upgrading to the head of BranchB after I run a downgrade of the latest revision on BranchB (using alembic upgrade heads).
Initially I thought it could be an issue with downgrade, but after some testing realised perhaps it the 'mis-use' of 'depends-on' on BranchB.
Hence, looking for advice here, in that should 'depends-on' only be used in certain circumstances? ie. when there is a need to run revision from BranchA before BranchB?

**Example (if applicable)**
The state of my alembic history:

```
BranchA: (base) revA1 ---> revA2 ---> revA3 ---> revA4 ---> revA5 ---> revA6
                                        \                                \
                                         \ (revB2 depends_on revA3)       \ (revB3 depends_on revA6)
                                          \                                \
BranchB: (base) revB1 -------------------------> revB2 --------------------------> revB3 ---> revB4 (with depends_on=revA6)
```

When I run 'alembic upgrade heads', I find 'alembic_version' table has only 'revB4'.
Then I add a new revision 'revB5' on top of 'revB4' with SAME depends on BranchA (ie. revA6) :

```
BranchA: (base) revA1 ---> revA2 ---> revA3 ---> revA4 ---> revA5 ---> revA6
                                        \                                \
                                         \ (revB2 depends_on revA3)       \ (revB3 depends_on revA6)
                                          \                                \
BranchB: (base) revB1 -------------------------> revB2 --------------------------> revB3 ---> revB4 --> revB5 (with depends_on=revA6)
```

On running 'alembic upgrade heads', the 'alembic_version' table has 'revB5' only.

Now when I run 'alembic downgrade BranchB@revB4 , the alembic_version table has 'revB4' AND 'revA6'. But this is not same as 'before', as it only had 'revB4'.

Now any attempt to upgade (alembic upgrade heads) fails with:

  ERROR [alembic.util.messaging] Requested revision revA6 overlaps with other requested revisions revB5
    FAILED: Requested revision revA6 overlaps with other requested revisions revB5

The only way I can make this work is manually delete 'revA6', then the upgrade works as per before.

[...]

PR #790, opened 2021-02-03
New downgrade algorithm to fix branch behaviour

simonbowly:
### Description

New algorithm for calculating downgrades which tries to be more systematic:

1. Parse branch@rev+rel syntax (and subsets) to determine which revision the command refers to.
2. Find immediate children of this revision, filter them using the branch label if necessary.
3. Find set of revisions to be dropped (result of step 2 + anything dependent on them), return them in reverse topologically sorted order (including dependencies) to ensure a valid downgrade path.

### To Review/Clairify

- 6075239 ignores order of revisions in downgrade tests to show they pass - this test change should probably be dropped but the order of revisions is hard to make deterministic?
* 3063264 changes tests from fixes to #789 - it seemed too many revisions were being dropped?
* What is the correct error message in failing test RevisionPathTest::test_invalid_relative_downgrade_path?
* Test cases + error messages needed for some asserts not currently hit
* Correct behaviour of `downgrade -1` when there are multiple heads (see xfailed tests in tests_version_traversal.py)?  This is documented but the order is ambiguous.  New implementation raises an error.

### Checklist

This pull request is:

- [ ] A short code fix
     - New tests added to better define downgrade behaviour when branches are involved

Apologies @zzzeek for leaving this hanging so long!

### DIFF
--- a/alembic/script/revision.py
+++ b/alembic/script/revision.py
@@ -771,6 +771,15 @@ class RevisionMap(object):
 
         """
 
+        if select_for_downgrade:
+            return self._iterate_revisions(
+                upper,
+                lower,
+                inclusive=inclusive,
+                implicit_base=implicit_base,
+                select_for_downgrade=select_for_downgrade,
+            )
+
         relative_upper = self._relative_iterate(
             upper,
             lower,
@@ -886,6 +895,122 @@ class RevisionMap(object):
                         )
                     )
 
+    def topological_sort(self, revisions, reverse=False):
+        """Yield revision ids of a collection of Revision objects in
+        topological sorted order (i.e. revisions always come after their
+        down_revisions and dependencies)."""
+        allitems = [d.revision for d in revisions]
+        edges = [
+            (rev, child.revision)
+            for child in revisions
+            if child.down_revision is not None
+            for rev in util.to_tuple(child.down_revision)
+        ] + [
+            (parent, child.revision)
+            for child in revisions
+            if child.dependencies is not None
+            for parent in util.to_tuple(child.dependencies)
+        ]
+        return sqlautil.topological.sort(edges, allitems)
+
+    def walk_down(self, start, steps, label):
+        """ Walk down the tree along a single path. """
+        assert steps <= 0
+        if isinstance(start, compat.string_types):
+            start = self.get_revision(start)
+        if steps == 0:
+            return start
+        if start is None:
+            raise ValueError("Can't walk back past base.")
+        children = self.get_revisions(start.down_revision)
+        if len(children) == 0:
+            raise util.CommandError(
+                f"Relative revision {label} didn't produce "
+                f"{abs(int(label))} migrations"
+            )
+        assert len(children) == 1
+        return self.walk_down(next(iter(children)), steps + 1, label)
+
+    def walk(self, start, steps, branch_label):
+        """ Walk up the tree along a single path. """
+        assert steps >= 0
+        if isinstance(start, compat.string_types):
+            start = self.get_revision(start)
+        if steps == 0:
+            return start
+        if start is None:
+            roots = [
+                rev
+                for rev in self._revision_map.values()
+                if rev is not None and rev.down_revision is None
+            ]
+            assert len(roots) == 1
+            return self.walk(roots[0], steps - 1, branch_label)
+        children = [
+            rev
+            for rev in self.get_revisions(start.nextrev)
+            if branch_label is None or branch_label in rev.branch_labels
+        ]
+        assert len(children) == 1
+        return self.walk(next(iter(children)), steps - 1, branch_label)
+
+    def _drop_inclusive(self, branch_revision, upper):
+        # Aim then is to drop :branch_revision; to do so we also need
+        # to drop its descendents and anything dependent on it.
+        drop_revisions = set(
+            self._get_descendant_nodes(
+                branch_revision,
+                include_dependencies=True,
+                omit_immediate_dependencies=False,
+            )
+        )
+        # Set logic/walking full tree might get expensive?
+        active_revisions = set(
+            self._get_ancestor_nodes(
+                self.get_revisions(upper), include_dependencies=True
+            )
+        )
+
+        # Emit revisions to drop in reverse topological sorted order.
+        drop_revisions = drop_revisions.intersection(active_revisions)
+
+        if len(drop_revisions) == 0:
+            # Empty intersection: target revs are not present.
+            raise RangeNotAncestorError(None, upper)
+
+        drop_reverse_topo_sorted = list(
+            reversed(list(self.topological_sort(drop_revisions)))
+        )
+        yield from self.get_revisions(drop_reverse_topo_sorted)
+
+    def _parse_downgrade(self, upper, lower):
+        # Parse lower: target revision + branch labels.
+        if lower is None:
+            return None, None
+        if isinstance(lower, compat.string_types):
+            match = _relative_destination.match(lower)
+            if match:
+                branch_label, symbol, relative = match.groups()
+                rel_int = int(relative)
+                if rel_int >= 0:
+                    return branch_label, self.walk(
+                        symbol, rel_int, branch_label
+                    )
+                else:
+                    if symbol is None:
+                        uppers = util.to_tuple(upper)
+                        assert len(uppers) == 1
+                        symbol = uppers[0]
+                    return branch_label, self.walk_down(
+                        symbol, rel_int, relative
+                    )
+            elif "@" in lower:
+                branch_label, _, symbol = lower.partition("@")
+                return branch_label, self.get_revision(symbol)
+            else:
+                return None, self.get_revision(lower)
+        raise ValueError(f"Failed to parse downgrade input '{lower}'")
+
     def _iterate_revisions(
         self,
         upper,
@@ -900,6 +1025,44 @@ class RevisionMap(object):
         across branches as a whole.
 
         """
+
+        if select_for_downgrade:
+
+            branch_label, target_revision = self._parse_downgrade(upper, lower)
+            # Find candidates to drop.
+            if target_revision is None:
+                # Downgrading back to base: find all tree roots.
+                roots = [
+                    rev
+                    for rev in self._revision_map.values()
+                    if rev is not None and rev.down_revision is None
+                ]
+            else:
+                # Downgrading to fixed target: find all direct children.
+                roots = list(self.get_revisions(target_revision.nextrev))
+
+            if branch_label and len(roots) > 1:
+                # Need to filter roots.
+                ancestors = {
+                    rev.revision
+                    for rev in self._get_ancestor_nodes(
+                        [self._resolve_branch(branch_label)],
+                        include_dependencies=False,
+                    )
+                }
+                # Intersection gives the root revisions we are trying to
+                # rollback with the downgrade.
+                roots = list(
+                    self.get_revisions(
+                        {rev.revision for rev in roots}.intersection(ancestors)
+                    )
+                )
+
+            # Ensure we didn't throw everything away.
+            assert len(roots) > 0, "No revisions identified to downgrade."
+
+            yield from self._drop_inclusive(roots, upper)
+            return
 
         requested_lowers = self.get_revisions(lower)
 
--- a/tests/test_version_traversal.py
+++ b/tests/test_version_traversal.py
@@ -310,12 +310,8 @@ class BranchedPathTest(MigrationTest):
             set([self.c2.revision]),
         )
 
-    @pytest.mark.xfail
     def test_relative_downgrade_baseplus2(self):
-        """Fails: alembic.util.exc.CommandError: Can't locate revision
-        identified by 'base'; even though it can find 'base' as a revision
-        marker on its own or with @branch syntax"""
-        # Fail
+        """ base+2 points to b, no branch label, drop everything above b. """
         self._assert_downgrade(
             "base+2",
             [self.d2.revision, self.d1.revision],
@@ -328,23 +324,12 @@ class BranchedPathTest(MigrationTest):
             set([self.b.revision]),
         )
 
-    @pytest.mark.xfail
     def test_relative_downgrade_branchplus2(self):
-        """Fails: goes one revision too far forward? Currently behaves as
-        expected for base+3."""
-        # Previous passing behaviour (downgrades both branches but leaves one stray)
-        # self._assert_downgrade(
-        #     "c2branch@base+2",
-        #     [self.d2.revision, self.d1.revision],
-        #     [
-        #         self.down_(self.d2),
-        #         self.down_(self.c2),
-        #         self.down_(self.d1),
-        #     ],
-        #     set([self.c1.revision]),
-        # )
-        # Correct behaviour (per https://github.com/sqlalchemy/alembic/pull/763#issuecomment-738741297)
-        # Only the c2branch should be downgraded, right back to base+2 = b
+        """
+        Correct behaviour (per
+        https://github.com/sqlalchemy/alembic/pull/763#issuecomment-738741297)
+        Only the c2branch should be downgraded, right back to base+2 = b
+        """
         self._assert_downgrade(
             "c2branch@base+2",
             [self.d2.revision, self.d1.revision],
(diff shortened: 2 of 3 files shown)


zzzeek:
wow this is huge ( in a good way).    My immediate reaction is, this is all in addition to what's already there, e.g. nothing is removed, I'd want to see if we can remove some of the complexity that was there already, because it's code that I can't even follow (my own code).    anyway lets put this through CI.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 60752399336b883d26f38d61a8337dd6fdd48087 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 60752399336b883d26f38d61a8337dd6fdd48087: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

simonbowly:
> wow this is huge ( in a good way). My immediate reaction is, this is all in addition to what's already there, e.g. nothing is removed, I'd want to see if we can remove some of the complexity that was there already, because it's code that I can't even follow (my own code). anyway lets put this through CI.

Cool - yes I ended up bypassing the old `_iterate_revisions` algorithm for downgrades.  The logic for upgrades could likely be replaced in the same way (new target heads -> find difference to current tree -> topo-sort), then there would be quite a bit of dead code to get rid of.

zzzeek:
OK so I have a lot to comment on, one immediate thing is that we are still supporting python 2 for 1.5 series :( so that changes some things right off.    Then the next thing is I think the walk() and walk_down() methods can be converted to use a while loop instead of recursion ?    python doesn't have tail recursion so these will overflow right now for a too-long list if I am reading them correctly.

This is really great and I'm just like starting to change everything around but I should let you iterate a bit.    If you were interested in using Gerrit directly we could both push changes (can PRs do that now?)

simonbowly:
No worries, I'll fix the string formats for python2 and replace the recursions later today, should be straightforward.  Happy to go through Gerrit - feel free to start making any changes you need.

alembic/script/revision.py
```diff
@@ -771,6 +771,15 @@ def iterate_revisions(
 
         """
 
+        if select_for_downgrade:
```
sqla-tester, line 774: **mike bayer** (zzzeek) wrote:

I think _iterate_revisions() might be better off split up into an _iterate_downgrade_revisions() / /_iterate_upgrade_revisions() for now as it seems to be split among a big conditional.

Otherwise this call to _iterate_revisions looks the same as the one at the bottom and the "select_for_downgrade" conditional can be reversed so that we only have to state _iterate_revisions once.   But I think for the moment splitting _iterate_revisions into two would be better.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
```diff
@@ -886,6 +895,122 @@ def _iterate_related_revisions(self, fn, targets, map_, check=False):
...
+        assert steps <= 0
+        if isinstance(start, compat.string_types):
+            start = self.get_revision(start)
+        if steps == 0:
+            return start
+        if start is None:
+            raise ValueError("Can't walk back past base.")
+        children = self.get_revisions(start.down_revision)
+        if len(children) == 0:
+            raise util.CommandError(
```
sqla-tester, line 927: **mike bayer** (zzzeek) wrote:

still supporting python 2.7:

        if len(children) == 0:
            raise util.CommandError(
                "Relative revision %(label)s didn't produce "
                "%(abslabel)s  migrations" % {
                    "label": label,
                    "abslabel": abs(int(label))
                }
            )

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
```diff
@@ -886,6 +895,122 @@ def _iterate_related_revisions(self, fn, targets, map_, check=False):
...
+                f"Relative revision {label} didn't produce "
+                f"{abs(int(label))} migrations"
+            )
+        assert len(children) == 1
+        return self.walk_down(next(iter(children)), steps + 1, label)
```
sqla-tester, line 932: **mike bayer** (zzzeek) wrote:

if im following correctly, if we had 100 children to traverse this would throw a recursion error?   any reason this function couldn't use a while loop and not call out to itself?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
```diff
@@ -886,6 +895,122 @@ def _iterate_related_revisions(self, fn, targets, map_, check=False):
...
+            ]
+            assert len(roots) == 1
+            return self.walk(roots[0], steps - 1, branch_label)
+        children = [
+            rev
+            for rev in self.get_revisions(start.nextrev)
+            if branch_label is None or branch_label in rev.branch_labels
+        ]
+        assert len(children) == 1
+        return self.walk(next(iter(children)), steps - 1, branch_label)
```
sqla-tester, line 955: **mike bayer** (zzzeek) wrote:

same idea here...

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
```diff
@@ -886,6 +895,122 @@ def _iterate_related_revisions(self, fn, targets, map_, check=False):
...
+        drop_revisions = drop_revisions.intersection(active_revisions)
+
+        if len(drop_revisions) == 0:
+            # Empty intersection: target revs are not present.
+            raise RangeNotAncestorError(None, upper)
+
+        drop_reverse_topo_sorted = list(
+            reversed(list(self.topological_sort(drop_revisions)))
+        )
+        yield from self.get_revisions(drop_reverse_topo_sorted)
```
sqla-tester, line 984: **mike bayer** (zzzeek) wrote:

we're still supporting python 2.7 for now, can't use this syntax

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
```diff
@@ -886,6 +895,122 @@ def _iterate_related_revisions(self, fn, targets, map_, check=False):
...
+
+    def _parse_downgrade(self, upper, lower):
```
sqla-tester, line 986: **mike bayer** (zzzeek) wrote:

is this doing what _relative_iterate does?  can these be unified?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

tests/test_version_traversal.py
```diff
@@ -308,15 +313,139 @@ def test_relative_upgrade(self):
...
+        )
+        self._assert_downgrade(
+            "c2branch@base+3",
+            [self.d2.revision, self.d1.revision],
+            [self.down_(self.d2)],
+            set([self.d1.revision, self.c2.revision]),
+        )
+
+    # Old downgrade -1 behaviour depends on order of branch upgrades.
+    # This should probably fail (ambiguous) but is currently documented
```
sqla-tester, line 362: **mike bayer** (zzzeek) wrote:

OK, I would assume this refers to the "downgrade -1" example at https://alembic.sqlalchemy.org/en/latest/branches.html.   So, can I assume that prior to this behavior, the example in the docs is basically random, right?   OK.   I see the failure here looks like this:

    File "/home/classic/dev/alembic/alembic/script/revision.py", line 1005, in _parse_downgrade
        assert len(uppers) == 1

OK great.  So, because we are at 1.5.2 / 1.5.3 etc. for this, we can't rip out that old behavior entirely right now.  Can we have your assertion run, and if it fails, emit a warning, that this case is deprecated, then proceed to have it do the old thing it was doing?   then we also need to update the docs to illustrate the "-1" thing using a proper branch label.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

tests/test_version_traversal.py
```diff
@@ -308,15 +313,139 @@ def test_relative_upgrade(self):
...
+    # as a key use case in branching.
+
+    @pytest.mark.xfail
```
sqla-tester, line 365: **mike bayer** (zzzeek) wrote:

we have our own decorators here:

    from alembic.testing import exclusions

    # ...

    @exclusions.fails()
    def test_thing(...):

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

sqla-tester (commented):

(9 comments)

as mentioned on the PR, very impressive!   I will have to spend a bunch of time studying and also trying to understand how the old version worked, unless you think you can rework this further to replace the old methods entirely.  right now with the "two totally different ways" I think things are too separate.

I will likely release Alembic 1.5.2 for now as this is a much bigger deal.   I'm almost wondering if this would be a 1.6 + python 3 only + we can break some backwards compat, but we *just* started the 1.5 series so it seems kind of soon.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

* alembic/script/revision.py (line 1075): do you think we could replace this part with the new approach?   because this code was always a nightmare.

zzzeek:
> No worries, I'll fix the string formats for python2 and replace the recursions later today, should be straightforward. Happy to go through Gerrit - feel free to start making any changes you need.

I'll wait for you to push up what you want to do with this next, and if you want to use gerrit I've added you to "contributors" so you sohuld be able to push changes directly.

I'm very impressed with what you've done here, I don't often get people getting into this level of intricacy so quickly so I am both glad to have more help and apologetic I wasn't sure if you were coming back :)

simonbowly:
Thanks!  Happy to be able to contribute - alembic has been a really useful tool for me :)

sqla-tester (commented):
**Simon Bowly** (simonbowly) wrote:

(9 comments)

Thanks for your comments! Fixed the minor ones. I think similar logic will work for upgrades if you want that old code replaced (downgrades entirely bypass it now)?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

* alembic/script/revision.py (line 1075): Yes I think we can use the same _parse_downgrade/_relative_iterate logic to get the target revisions for upgrade as well, then simplify things using the set logic approach. I'll give it go.

alembic/script/revision.py
sqla-tester: **Simon Bowly** (simonbowly) wrote:

Done - realised while doing this that I was ignoring some options.  Can you clarify behaviour of implicit_base=True?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
sqla-tester: **Simon Bowly** (simonbowly) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
sqla-tester: **Simon Bowly** (simonbowly) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
sqla-tester: **Simon Bowly** (simonbowly) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
sqla-tester: **Simon Bowly** (simonbowly) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

alembic/script/revision.py
sqla-tester: **Simon Bowly** (simonbowly) wrote:

Yes I think it is doing something similar. I'm struggling to follow _relative_iterate though; I think it's cleaner to avoid having two separate cases for relative and absolute revisions. Once you determine what's being pointed to (either using absolute or relative syntax) the logic for determining revisions to drop shouldn't depend on how the target was specified. Should be able to merge the two.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

tests/test_version_traversal.py
sqla-tester: **Simon Bowly** (simonbowly) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

tests/test_version_traversal.py
sqla-tester: **Simon Bowly** (simonbowly) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

sqla-tester:

hey there.   I'm sort of getting other things done and am semi-procrastinating diving into this for which I feel I would need to look at all my old code and figure out what I was thinking with all of that.   I've just released SQLAlchemy 1.4.0b2 and am trying to focus on getting things lined up for 1.4 / 2.0 etc.   If you are blocked on getting tests to pass here, let me know and I'll see if I can jump in sooner, otherwise the best goal we can have here would be to use your newer approach fully, assuming it does everything needed.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

sqla-tester:
**Simon Bowly** (simonbowly) wrote:

> Patch Set 2:
> 
> hey there.   I'm sort of getting other things done and am semi-procrastinating diving into this for which I feel I would need to look at all my old code and figure out what I was thinking with all of that.   I've just released SQLAlchemy 1.4.0b2 and am trying to focus on getting things lined up for 1.4 / 2.0 etc.   If you are blocked on getting tests to pass here, let me know and I'll see if I can jump in sooner, otherwise the best goal we can have here would be to use your newer approach fully, assuming it does everything needed.

No worries. I'm working on replacing the upgrade code with the same approach so no need to go through the old stuff if you want to avoid it. I have a few downgrade tests that are still failing but they should be manageable.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

sqla-tester:
**Simon Bowly** (simonbowly) wrote:

Okay, almost done with this I think, I'm just stuck on 4 failing test cases relating to dependencies on branch labels. Could you clarify how these are intended to work?

For test DependsOnBranchLabelTest.test_upgrade_path:

a1 -> b1 -> c1 (label c1_lib) -> d1 -> e1 -> f1
a2 -> b2 -> c2 (depends on c1_lib)

Does this indicate that a2,b2,c2 *all* depend on c1? This is what the test seems to require as if the revisions are not done in the order a1,b1,c1,b2,c2, the result is two heads rather than one (although, that could just be a problem with HeadMaintainer). On the other hand the test starts with head 'a2' but none of the c1_lib branch, which seems inconsistent?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

zzzeek:
> **Simon Bowly** (simonbowly) wrote:
> 
> Okay, almost done with this I think, I'm just stuck on 4 failing test cases relating to dependencies on branch labels. Could you clarify how these are intended to work?
> 
> For test DependsOnBranchLabelTest.test_upgrade_path:
> 
> a1 -> b1 -> c1 (label c1_lib) -> d1 -> e1 -> f1
> a2 -> b2 -> c2 (depends on c1_lib)
> 
> Does this indicate that a2,b2,c2 _all_ depend on c1? 

I haven't looked at the test but based on the above diagram, a2 and b2 should not depend on the "c1" label, just c2.

> This is what the test seems to require as if the revisions are not done in the order a1,b1,c1,b2,c2, the result is two heads rather than one (although, that could just be a problem with HeadMaintainer).

c1 and c2 at the same time?   I just fixed an issue with this kind of thing in #789 .   Assuming you're rebased onto master and have the changes from 63da35758c11fe5450f4a5b932ed8dd2fb559302 then...it may be another bug?   if further operations with the heads in that state can't continue, then that seems like a bug yes.

> On the other hand the test starts with head 'a2' but none of the c1_lib branch, which seems inconsistent?

if a2 doesnt depent on c1_lib then that should be fine.

>

sqla-tester:
**Simon Bowly** (simonbowly) wrote:

Thanks - I'd rebased onto your change but wasn't using _normalized_down_revisions. Passes now.
This is close to being done, have a fairly complete set of tests describing branch behaviour. Changed quite a few existing tests in the process so I hope they are correct?

Should doc additions be made in this PR or elsewhere?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2530 has been **merged**. Congratulations! :)

PR #834, opened 2021-04-30
Require setuptools, alembic/util/pyfiles.py, alembic/script/write_hooks.py, tests/test_post_write.py import pkg_resources

hrnciar:
### Description

List setuptools as runtime dependency.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -46,6 +46,7 @@ install_requires =
     Mako
     python-editor>=0.3
     python-dateutil
+    setuptools
 
 [options.packages.find]
 exclude =


CaselIT:
Hi, 

I don't think we are using setuptools as a dependency with using alembic.
Also I'm not really sure pythin is able to work without it, so it should be installed in the environment.

What's the use-case that fails with it as a missing dependency?

hrnciar:
In Fedora, we package RPM package for alembic. I am not the maintainer of that package, but the maintainer of Python in Fedora.

We used to automatically generate a runtime dependency on setuptools for all RPM packages with Python projects with console_scripts entrypoints. Reason for this was that it was necessary for the way how was entrypoint script generated. Nowadays the entrypoint script looks differently. Particularly, they do not require setuptools to be installed on runtime in order to execute (at least on Python 3.8+ -- because it has importlib.metadata).

Hence we are about to stop generating the automatic runtime dependency on setuptools for packages with console_scripts entrypoints. We went through all the affected packages in Fedora and identified some packages that will be negatively impacted in Fedora because when we stop adding the requirement automatically, they will have missing dependencies, because they use setuptools (pkg_resources) on runtime, but don't specify a dependency.

As long as you import from pkg_resources in the actual code, the project requires setuptools on runtime and the requirement needs to be specified, otherwise it might be missing.

CaselIT:
> In Fedora, we package RPM package for alembic. I am not the maintainer of that package, but the maintainer of Python in Fedora.

ok, so ideally one could have python installed then install alembic from rmp, all without installing pip.

makes sense I guess. cc @zzzeek

zzzeek:
OK to my total shock we have pkg_resources imports in alembic and I would vastly prefer we get rid of those.   i dont want setuptools as a dependency.

can replace with importlib.metadata here:

https://github.com/sqlalchemy/alembic/blob/master/alembic/script/write_hooks.py#L126

for this one, I'm not sure yet what we can do, but setuptools can at least be an optional dependency:

https://github.com/sqlalchemy/alembic/blob/master/alembic/util/pyfiles.py#L45

zzzeek:
agree we can close this and create new issue to remove pkg_resources dependency from codebase?   1.7 milestone

CaselIT:
@zzzeek we still need to do this right?

zzzeek:
yes

zzzeek:
I propose we close this and address via #885

CaselIT:
I think we replaced pkg_resources, so we can close.

Thanks for raising the issue

PR #873, opened 2021-07-12
Correct typo in documention, closes #872

### LINKED ISSUE #872: Typo in docs, `type` is not the same thing as `type_`
**Describe the bug**

At https://alembic.sqlalchemy.org/en/latest/autogenerate.html#omitting-table-names-from-the-autogenerate-process

There's a code snippet which reads:

```
target_metadata = MyModel.metadata

def include_name(name, type_, parent_names):
    if type == "schema":
        return name in [None, "schema_one", "schema_two"]
    elif type_ == "table":
        # use schema_qualified_table_name directly
        return (
            parent_names["schema_qualified_table_name"] in
            target_metadata.tables
        )
    else:
        return True

context.configure(
    # ...
    target_metadata = target_metadata,
    include_name = include_name,
    include_schemas = True
)
```

But the first line of that `include_name` function is incorrect, `type == "schema"` will always be false because `type` is a Python keyword.  I believe it should be:

```
def include_name(name, type_, parent_names):
    if type_ == "schema":
```

pzelnip:
### Description

Corrects a typo in the sample code at https://alembic.sqlalchemy.org/en/latest/autogenerate.html#omitting-table-names-from-the-autogenerate-process

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/docs/build/autogenerate.rst
+++ b/docs/build/autogenerate.rst
@@ -356,7 +356,7 @@ full example of omitting tables with schema support may look like::
     target_metadata = MyModel.metadata
 
     def include_name(name, type_, parent_names):
-        if type == "schema":
+        if type_ == "schema":
             return name in [None, "schema_one", "schema_two"]
         elif type_ == "table":
             # use schema_qualified_table_name directly


CaselIT:
Thanks!

Issue #879, opened 2021-07-30
Drop/CreateTableOp.from_table() no longer copy the table's `Schema.info`


cansjt:
**Describe the bug**

Since 1.6.0 it seems CreateTableOp.from_table() and DropTableOp.from_table() no longer copy the table's `Schema.info` dictionary. The code of the methods substantially changed between 1.5.8 and 1.6.0

**Expected behavior**

The `info` dictionary should still contain the same data as the one of the original table.

Not sure if other bits of the code could be impacted (other `SchemaItem` loosing their `info` data)

**To Reproduce**

```py
# Here is the shortest I could come up with
from sqlalchemy import create_engine, MetaData, Table
from sqlalchemy.ext.compiler import compiles
from sqlalchemy.schema import DropTable

from alembic.migration import MigrationContext
from alembic.operations import Operations
from alembic.operations.ops import DropTableOp


@compiles(DropTable)
def _add_if_exists(element, compiler, **kw):
    """Patches DROP TABLE statements to add IF EXISTS clause

    Shamelessly stolen from https://github.com/sqlalchemy/alembic/issues/151
    """
    assert element.element.info == THE_TABLE.info
    output = compiler.visit_drop_table(element, **kw)
    # if isinstance(element.element.info, dict) and element.element.info.get('ifexists'):
    #     output = re.sub(
    #         r"^\s*DROP TABLE", "DROP TABLE IF EXISTS", output, re.S)
    return output


THE_TABLE = Table('name', MetaData(), info={'ifexists': True})
operation = DropTableOp.from_table(THE_TABLE)
engine = create_engine('sqlite://')
with engine.connect() as connection:
    ctx = MigrationContext.configure(connection)
    operations = Operations(ctx)
    operations.invoke(operation)
```

Works fine with Alembic up to 1.5.8

Context:

**Error**

```
Traceback (most recent call last):
  File "minimal.py", line 37, in <module>
    operations.invoke(operation)
  File "~/.virtualenvs/dtp38/lib/python3.8/site-packages/alembic/operations/base.py", line 354, in invoke
    return fn(self, operation)
  File "~/.virtualenvs/dtp38/lib/python3.8/site-packages/alembic/operations/toimpl.py", line 72, in drop_table
    operations.impl.drop_table(
  File "~/.virtualenvs/dtp38/lib/python3.8/site-packages/alembic/ddl/impl.py", line 297, in drop_table
    self._exec(schema.DropTable(table))
[...]

Issue #849, opened 2021-05-25
`DropIndexOp` does not render keyword arguments such as `postgresql_concurrently`


jetzhou:
**Describe the bug**
https://github.com/sqlalchemy/alembic/commit/101136605150d50fe0a3b6e7bc6c2ad90bbeb9df added kwargs support to drop index operation in SQL generation. However, in autogenerate render step, an operation such as this one:
```py
alembic.operations.ops.DropIndexOp(
    'example_index',
    postgresql_concurrently=True,
)
```
will be rendered to
```py
op.drop_index(op.f('example_index'))
```
which ignores the `postgresql_concurrently` kwarg.

**Expected behavior**
```py
alembic.operations.ops.DropIndexOp(
    'example_index',
    postgresql_concurrently=True,
)
```
should render in autogenerate to 
```py
op.drop_index(op.f('example_index'), postgresql_concurrently=True)
```

**To Reproduce**
Test code that does not pass
```py
    def test_render_drop_index_custom_kwarg(self):
        t = Table(
            "test",
            MetaData(),
            Column("id", Integer, primary_key=True),
            Column("active", Boolean()),
            Column("code", String(255)),
        )
        idx = Index(None, t.c.active, t.c.code, somedialect_foobar="option")
        op_obj = ops.DropIndexOp.from_index(idx)
        eq_ignore_whitespace(
            autogenerate.render_op_text(self.autogen_context, op_obj),
            "op.drop_index(op.f('ix_test_active'), table_name='test', "
            "somedialect_foobar='option')",
        )

```

**Versions.**
 - OS: Ubuntu 21.04 
 - Python: 3.9.5
 - Alembic: 1.6.4
 - SQLAlchemy: 1.3.19
 - Database: PostgreSQL
 - DBAPI: psycopg

**Additional context**
Proposed fix: https://github.com/jetzhou/alembic/commit/6392a287179ed746f709ba1e0e07ccab3ea8e4c6

**Have a nice day!**

zzzeek:
will have to look and see if we have existing precedent for this, which I think we should, like for indexes etc.

zzzeek:
looks like you have  a fix ready!  that's a PR, please send.    I assume most of the other constructs are OK in this regard

jetzhou:
Thanks @zzzeek for the triage! Created PR for this.

zzzeek:
thanks for following all the guidelines on this!   most people dont :)  but it makes it easy

[...]

PR #881, opened 2021-07-31
Preserve comment and info in Drop/CreateTableOp

cansjt:
This is a fix for issue #879 

### Description

The `DropTableOp.from_table()` method loses the table's `SchemaItem.info` and `comment` attributes while the `CreateTableOp.from_table()` only loses the `info` attribute.

This means the following assertion does not hold, when a table goes through a from_table() to_table() round trip:
```
assert table == Op.from_table(table).to_table()  # Symbolic (see note below)
```
This commit fixes both issues.

Notes:

1) While fixing this I found the implementations of the Create/DropTableOp to be strangely assymetrical, in particular wrt. the way each deals with extra keywords one can pass to a table. Happy to refactor this a little if you'd like me to.

2) There is table equality notion implemented. But it seems reliant on the identity of the tables `MetaData` though, which kind of makes sense if you want to be 100% it is the same object, in the exact same database (given the DB URI+schema name+table name, etc.).
```
>>> from sqlalchemy import Table, MetaData
>>> t1 = Table('name', MetaData(schema='plop'))
>>> t2 = Table('name', MetaData(schema='plop'))
>>> t1 == t2
False
>>> t3 = Table('name', t1.metadata)
>>> t1 == t3
True
```
Given that the `MetaData` object is going "bind-free" in SQLAlchemy 2.0, I wonder if this definition of table equality is still relevant, though.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

Thanks, you too !


zzzeek:
hello there -

thanks for this!  has a test and everything.  I am on vacation this week so have limited time to check into things.   will run this through CI.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 6ed4ed0f65d0273ac15402e7b65403c7c7285a14 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 6ed4ed0f65d0273ac15402e7b65403c7c7285a14: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

alembic/operations/ops.py
```diff
@@ -959,6 +959,7 @@ def __init__(
         self.comment = kw.pop("comment", None)
         self.prefixes = kw.pop("prefixes", None)
         self.kw = kw
+        self.table_info = kw.pop('info', {})
```
sqla-tester, line 962: **mike bayer** (zzzeek) wrote:

call this ".info"

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

alembic/operations/ops.py
```diff
@@ -1091,10 +1094,18 @@ def create_table(cls, operations, table_name, *columns, **kw):
 class DropTableOp(MigrateOperation):
     """Represent a drop table operation."""
 
-    def __init__(self, table_name, schema=None, table_kw=None, _reverse=None):
+    def __init__(self,
+                 table_name,
+                 schema=None,
+                 table_comment=None,
```
sqla-tester, line 1100: **mike bayer** (zzzeek) wrote:

I think these both need to come from table_kw using the "pop" form.  if we were to refactor this further, "table_kw" would become "**kw" just like CreateTableOp.

Also we need to add "prefixes" here.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

alembic/operations/ops.py
```diff
@@ -1091,10 +1094,18 @@ def create_table(cls, operations, table_name, *columns, **kw):
...
+                 table_info=None,
+                 table_kw=None,
+                 _reverse=None):
         self.table_name = table_name
         self.schema = schema
+        self.comment = table_comment
         self.table_kw = table_kw or {}
+        self.table_info = table_info or {}
```
sqla-tester, line 1108: **mike bayer** (zzzeek) wrote:

implement as:

  self.info = table_kw.pop("info", None)

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

tests/test_op.py
```diff
@@ -1152,9 +1152,15 @@ def test_create_index_add_kw(self):
...
-        table = schema_obj.table("x", Column("q", Integer))
+        table = schema_obj.table("x",
+                                 Column("q", Integer),
+                                 info={'custom': 'value'},
+                                 postgresql_partition_by='x',
+                                 comment='some comment')
         op = ops.DropTableOp.from_table(table)
         is_not_(op.to_table(), table)
+        eq_(op.to_table().comment, table.comment)
+        eq_(op.to_table().info, table.info)
```
sqla-tester, line 1164: **mike bayer** (zzzeek) wrote:

see if we can add "prefixes"

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

tests/test_op.py
```diff
@@ -1171,9 +1177,15 @@ def test_drop_table_add_kw(self):
...
-        table = schema_obj.table("x", Column("q", Integer))
+        table = schema_obj.table("x",
+                                 Column("q", Integer),
+                                 postgresql_partition_by='x',
+                                 info={'custom': 'value'},
+                                 comment='some comment')
         op = ops.CreateTableOp.from_table(table)
         is_not_(op.to_table(), table)
+        eq_(op.to_table().comment, table.comment)
+        eq_(op.to_table().info, table.info)
```
sqla-tester, line 1189: **mike bayer** (zzzeek) wrote:

see if we can add "prefixes"

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

sqla-tester (commented):

hi and thanks for doing this!

changes we need include:

1. a changelog file docs/build/unreleased/879.rst needs to be added
2. we do need some factoring as you mention to make DropTableOp look like CreateTableOp but I think it should be for now that CreateTableOp->**kw and DropTableOp->table_kw are treated equivalently for now, no new arguments added
3. for pep8/flake8, black formatting needs to be run on the files.   Best way to do this is to use pre-commit:

   $ cd /path/to/alembic
   $ pre-commit install

the correct formatting steps will take place when you next commit.    do a "git commit --amend" and force-push the PR to make sure it gets all the files that were changed.

im on vacation this week so i may not respond quickly. 

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

* /COMMIT_MSG (line 24): that's public API and while I can see DropTable is different, that's because the arguments sent to Table are currently only used during CREATE TABLE; I believe table_kw was added to better support "reversal" of the DropTableOp back into CreateTableOp.   would not change any of that within this change.
* /COMMIT_MSG (line 36): this definition of "==" is still relevant; if you want a comparison of two Table objects based on their contents, use the .compare() method, 1.4 and above:

    >>> from sqlalchemy import Table, MetaData
    >>> t1 = Table('name', MetaData(schema='plop'))
    >>> t2 = Table('name', MetaData(schema='plop'))
    >>> t1.compare(t2)
    True
    >>> t3 = Table('notname', MetaData(schema='plop'))
    >>> t1.compare(t3)
    False
* /COMMIT_MSG (line 62): please add:

    Fixes: #879

alembic/operations/ops.py
```diff
@@ -1108,7 +1114,10 @@ def from_table(cls, table, _namespace_metadata=None):
         return cls(
             table.name,
             schema=table.schema,
-            table_kw=table.kwargs,
+            comment=table.comment,
+            info=table.info,
+            prefixes=table._prefixes,
```
cansjt, line 1119: Given that this is a list and is hence mutable. shouldn't we copy it ? Same for info, which is a dict.

alembic/operations/ops.py
```diff
@@ -1108,7 +1114,12 @@ def from_table(cls, table, _namespace_metadata=None):
         return cls(
             table.name,
             schema=table.schema,
-            table_kw=table.kwargs,
+            table_kw={
+                "comment": table.comment,
+                "info": table.info,
+                "prefixes": table._prefixes,
+                **table.kwargs,
+            },
```
cansjt, line 1122: This is the only way I can see to not break the class' `__init__()` public API.

I cannot help but wonder how much of public API this is though. Indeed, even when reverting a `CreateTableOp`, for example, `DropTableOp.from_table()` is used (see l. 967). But on principle I agree, we should not break it.

alembic/operations/ops.py
```diff
@@ -1094,7 +1097,10 @@ class DropTableOp(MigrateOperation):
     def __init__(self, table_name, schema=None, table_kw=None, _reverse=None):
```
cansjt, line 1097:
```suggestion
    def __init__(self, table_name, schema=None, table_kw=None, _reverse=None, **kw):
        if table_kw:
            warnings.warn('...', DeprecationWarning)
            kw.update(table_kw)
        self.table_name = table_name
        self.schema = schema
        self.comment = kw.pop('"comment", None)
        self.info = kw.pop('"info", None)
        self.prefixes = kw.pop("prefixes", None)
        self.table_kw = kw
```
Would something like it help transition ? I don't like it much, because it would allow to pass twice the same "kwargs", with different values, once through `table_kw`, once though the regular `**kw`.
I would at least add a check that there is no override. It would also requite to fix the `drop_table()` class method.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 257b987e5b881da1704be4f72674f01609864826 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 257b987e5b881da1704be4f72674f01609864826 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

alembic/operations/ops.py
zzzeek: I guess this is a good idea, seems to not be critical but then i may not be envisioning how these structures may be used, so sure.

alembic/operations/ops.py
zzzeek: we are on the edge of a new "minor" series 0.7 so it's a better time to change the API but I felt for the moment it's best to make this PR work without changing any public API, then make public API changes between these two objects something separate.

alembic/operations/ops.py
zzzeek: i wouldnt want them both at the same time for the moment.    again I want to get the feature "working" and then if we want to consolidate/clean up/use both/one or the other of "kw" / "table_kw" i want to look at that separately.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision e509b9e8180085998e32746721eeb9266b9f2145 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset e509b9e8180085998e32746721eeb9266b9f2145 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

zzzeek:
specifically, i want to look at the `**kw` situation for all the operations objects, like indexes and all that.  every object has dialect-specific kw so i want to apply one consistent approach to the whole thing and do real deprecation etc.

sqla-tester:

this looks good to me as is, im going to add a changelog file and also revert that formatting change for now, if we want to work with the kw API to make it make more sense we can approach that separately.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/2989

zzzeek:
also #849 was a similar issue

zzzeek:
oh i guess black has made those formatting changes, OK.

zzzeek:
i will also dict-/list- ify info and prefixes.

cansjt:
I guess I won't... couldn't push to github.

zzzeek:
yeah i wouldnt use deepcopy() here.   this is pushed.  thanks!

PR #926, opened 2021-09-18
Reorder imports to be more linter friendly

abitrolly:


alembic/templates/generic/script.py.mako
```diff
@@ -5,8 +5,9 @@ Revises: ${down_revision | comma,n}
 Create Date: ${create_date}
 
 """
-from alembic import op
```
zzzeek, line 8: this is fine but would need to be applied to all the script.py.mako templates in all the configs.

zzzeek (changes requested):
needs to be applied consistently.

abitrolly:
Found 4 `script.py.mako` files. That should be it.

abitrolly:
Tested with `python -m isort --check migrations/versions`.

CaselIT (changes requested):
Personally I'm not convinced we need this, since in making isort happy, but probably will make other linters unhappy.

In any case, if we need to have it, there should not be a space between sqlalchemy and alembic, since both are considered as "packages" by isort

abitrolly:
https://github.com/pypa/warehouse uses `lines_between_types = 1` on their config, so my check did work without empty lines.

CaselIT:
It again reinforces my point that formatting is something alembic cannot solve since it depends on the users.

In any case the change is fine for me

CaselIT (approved):

alembic/templates/generic/script.py.mako
abitrolly: Done.

abitrolly:
@CaselIT you're right, but at least I've tried to improve the contributor experience.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 035b9513c3ccf9dd22e21024f52ddba9a3730548 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 035b9513c3ccf9dd22e21024f52ddba9a3730548: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/3093

sqla-tester:

hi -

propose abandon this change as it makes no sense.  There is no standard "import order" and the proposed ordering fails the import ordering that the SQLAlchemy and Alembic projects themselves use.  Please set up as follows:

setup.cfg

[flake8]
show-source = false
enable-extensions = G

exclude = .venv,.git,.tox,dist,doc,*egg,build
import-order-style = google


tox.ini

[tox]
isolated_build = True

[testenv:pep8]
basepython = python3
deps=
      flake8
      flake8-import-order
commands =
     flake8 .



given two files with the two different imports here, the proposed one fails:


/mypackage/modelbase.py:2:1: I100 Import statements are in the wrong order. 'from alembic import op' should be before 'import sqlalchemy'


There is no fixed import order style in Python, and the post_process hooks are designed for this use case.



View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/3093

sqla-tester:

the existing ordering is more correct

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/3093

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/3093 has been **abandoned**.  That means that at least for the moment I need to close this pull request. Sorry it didn't work out :(

abitrolly:
@zzzeek in the discussion https://github.com/sqlalchemy/alembic/discussions/925#discussioncomment-1351237 you say it is OK, and during the review you say it makes no sense. What happened?

I don't really plan to invest more time into pushing this change. My experience shows that if any action meets a counter-action, which I can not explain, it is better to just abandon it. Open source is fun to people, because it allows to be opinionated, and if I don't like anything, I could just fork it.

Having said that I will repeat my opinion for the reference.

1. If Alembic uses unusual formatting rules for own code, it should use the config rather than imposing these conflicting rules to projects that prefer the most popular no-config linting consensus.
2. `black` became popular not because other linters didn't exist, were badly written or limited.

CaselIT:
I honestly still don't understand what you want us to do..

>  If Alembic uses unusual formatting rules for own code, it should use the config rather than imposing these conflicting rules to projects that prefer the most popular no-config linting consensus.

why is it relevant how alembic own code is formatted? And what is the most popular linting consensus? where is it defined?
regarding no-config, alembic is requires it and is not able to work without a configuration, so I don't think that fits.

> black became popular not because other linters didn't exist, were badly written or limited.

not sure how's that relevant. alembic is not trying to be a formatter..

abitrolly:
> I honestly still don't understand what you want us to do..

It was an interesting discussion, that's it.

abitrolly:
> why is it relevant how alembic own code is formatted?

Because it was one of the arguments to close this PR.

abitrolly:
> And what is the most popular linting consensus? where is it defined?

It is not defined. It is measured. I explained one way to measure it in https://github.com/PyCQA/isort/issues/1812 The UX discipline studies such things in a greater detail.

abitrolly:
> not sure how's that relevant. alembic is not trying to be a formatter..

Yes, I should have expanded the thought to be complete. The popularity of the `black` is a proof that convenient defaults matter.

PR #929, opened 2021-09-22
Fix issue 928 create_check_constraint condition argument typing fix

liverpool1026:
### Description
The typing of condition argument is incorrect according to the documentation.
Definition:
create_check_constraint(constraint_name: Optional[str], table_name: str, condition: BinaryExpression, schema: Optional[str] = None, **kw) → Optional[Table]

The definition specifies the type for condition must be BinaryExpression, but in the documentation below.

condition¶ – SQL expression that’s the condition of the constraint. Can be a string or SQLAlchemy expression language structure.

### Checklist
This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


CaselIT:
I have to look better, but I think we also accept strings.

cc @zzzeek this is supported right?
```py

    op.create_check_constraint(
        "check_constraint",
        "table_name",
        "applies_from_time < applies_to_time"
    )
```

zzzeek:
yes im sure we support that

liverpool1026:
> I have to look better, but I think we also accept strings.
> 
> cc @zzzeek this is supported right?
> 
> ```python
>     op.create_check_constraint(
>         "check_constraint",
>         "table_name",
>         "applies_from_time < applies_to_time"
>     )
> ```

Yeah, it was explicitly mentioned in the doc string that string is allowed.

PR #969, opened 2021-11-18
Tweak docs about version_path_separator

ziima:
I tried to improve docs regarding `version_path_separator` based on my experience with it.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/templates/generic/alembic.ini.mako
+++ b/alembic/templates/generic/alembic.ini.mako
@@ -35,16 +35,18 @@ prepend_sys_path = .
 # version location specification; This defaults
 # to ${script_location}/versions.  When using multiple version
 # directories, initial revisions must be specified with --version-path.
-# The path separator used here should be the separator specified by "version_path_separator"
+# The path separator used here should be the separator specified by "version_path_separator" below.
 # version_locations = %(here)s/bar:%(here)s/bat:${script_location}/versions
 
 # version path separator; As mentioned above, this is the character used to split
-# version_locations. Valid values are:
+# version_locations. The default within new alembic.ini files is "os", which uses os.pathsep.
+# If this key is omitted entirely, it falls back to the legacy behavior of splitting on spaces and/or commas.
+# Valid values for version_path_separator are:
 #
 # version_path_separator = :
 # version_path_separator = ;
 # version_path_separator = space
-version_path_separator = os  # default: use os.pathsep
+version_path_separator = os  # Use os.pathsep. Default configuration used for new projects.
 
 # the output encoding used when revision files
 # are written from script.py.mako
--- a/docs/build/branches.rst
+++ b/docs/build/branches.rst
@@ -521,11 +521,12 @@ that module.  So to start out, we can edit ``alembic.ini`` to refer
 to multiple directories;  we'll also state the current ``versions``
 directory as one of them::
 
+  # A separator for the location paths must be defined first.
+  version_path_separator = os  # Use os.pathsep.
   # version location specification; this defaults
   # to foo/versions.  When using multiple version
   # directories, initial revisions must be specified with --version-path
-  version_path_separator = space
-  version_locations = %(here)s/model/networking %(here)s/alembic/versions
+  version_locations = %(here)s/model/networking:%(here)s/alembic/versions
 
 The new directory ``%(here)s/model/networking`` is in terms of where
 the ``alembic.ini`` file is, as we are using the symbol ``%(here)s`` which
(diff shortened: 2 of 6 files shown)


alembic/templates/generic/alembic.ini.mako
```diff
@@ -35,16 +35,16 @@ prepend_sys_path = .
 # version location specification; This defaults
 # to ${script_location}/versions.  When using multiple version
 # directories, initial revisions must be specified with --version-path.
-# The path separator used here should be the separator specified by "version_path_separator"
+# The path separator used here should be the separator specified by "version_path_separator" below.
 # version_locations = %(here)s/bar:%(here)s/bat:${script_location}/versions
 
 # version path separator; As mentioned above, this is the character used to split
-# version_locations. Valid values are:
+# version_locations. It should be set, if multiple paths is used. Valid values are:
```
zzzeek, line 42: since we are looking to make this clearer, I dont understand this sentence, "it should be set, if multiple paths are used".   this value is always "set" to something even if it's not in the config file.   do you mean to say "it should be changed" ?   if so, what should it be changed to?    maybe instead a line like, "this should be changed to reflect the path separator used for the version_locations configuration variable".    otherwise this doesnt seem to make it any clearer.

zzzeek (changes requested):
hi there and thanks for this!

I agree we can make this verbiage better.   However I think we should make sure we understand what it was that you didn't find clear and then make sure that's clear, without introducing new ways to make it non-clear.    As written it looks like there is some way that the value can be "not" set, which doesnt make sense.

alembic/templates/generic/alembic.ini.mako
ziima: I admit, I did get somewhat confused by note that `os.pathsep` is default in docs, but it is actually a space in code. From the code I got the impression that `version_path_separator` unset is discouraged backward compatible case. Did I understand it incorrectly?

zzzeek: yes that is correct!     OK so the confusion you had was, "if omitted, defaults to "space", which is legacy"

so how about we make this more clear like this

```
# version path separator; As mentioned above, this is the character used to split
# version_locations.  The default within new alembic.ini files is "os", which uses
# os.pathsep.  If this key is omitted entirely, falls back to the legacy behavior
# of splitting on spaces and/or commas.
# Valid values for version_path_separator are:
#
# version_path_separator = :
# version_path_separator = ;
# version_path_separator = space
version_path_separator = os  # uses os.pathsep.  
```

ziima: OK, I fixed that.

Also, shouldn't the legacy version trigger a deprecation warning? I could add it in a separate MR.

CaselIT: I think it may make sense, what's your opinion @zzzeek ?

as a side note, do we have deprecations warnings on alembic?

ziima: I found this warning https://github.com/sqlalchemy/alembic/blob/main/alembic/util/langhelpers.py#L125-L128 which looks like a deprecation, although it's not marked as such.

zzzeek: we dont have deprecation warning mechanics set up in alembic , and I dont see that it's necessary to remove the "legacy" path separator style.  maybe it's cleaner just to call it "legacy" and have it be an entry in the config?    if we were to totally remove the element, what would the default be?

ziima: From the docs, it seemed to me that `os` is to be the new default.

zzzeek: it's the default, *in the config file*.   people that start new alembic projects will see this parameter is already in the config file.  so there's really two levels of "default" for this kind of thing in Alembic, there's, "what we have in the pre-fab templates" and then there's "what we do if the config value is missing entirely".    this should also be worked into whatever language we do here.

Really if we're going to depreacte the old version I would kind of want to raise an error if this parameter isn't in the config, because we want it to be explicit.

ziima: OK, I leave it be. Otherwise, this MR should be complete.

docs/build/branches.rst
```diff
@@ -521,11 +521,12 @@ that module.  So to start out, we can edit ``alembic.ini`` to refer
 to multiple directories;  we'll also state the current ``versions``
 directory as one of them::
 
+  # A separator for the location paths must be defined first.
+  version_path_separator = os  # Use os.pathsep.
```
zzzeek, line 525: nice

docs/build/tutorial.rst
```diff
@@ -307,6 +309,10 @@ This file contains the following features:
   allow revisions to exist in multiple directories simultaneously.
   See :ref:`multiple_bases` for examples.
 
+* ``version_path_separator`` - a separator of ``version_locations`` paths.
```
zzzeek, line 312: good catch

zzzeek (approved):
looks great thanks!

Issue #1027, opened 2022-05-02
Add epoch as an option for the file_template


lowercase00:
This is just a small detail, but would be great to have epoch as an alternative for the file_template, I'll be opening a PR with this change.

sqla-tester:
**Caio Carvalho** has proposed a fix for this issue in the **main** branch:

**Add epoch as an option for file_template** https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/3823

PR #1028, opened 2022-05-02
Add epoch as an option for file_template

lowercase00:
Hi all, as we are not huge fans of big file names, we would rather use epoch as a way to have the migrations in chronological order. I made a small change (with docs and test) to allow the epoch to be used in the `file_template`. From what I could see, this shouldn't have many side effects. As this is my first PR around here, please let me know if there's anything that I should do differently. 

### Description
- Added epoch as an option to the `file_template`.
- Added a test to make sure the epoch is considered on the name
- Added a test to make sure the epoch formula is correct (one extra second increases epoch by 1)

### Checklist

This pull request is:

- [ ] A short code fix
	- Fixes: #1027

### DIFF
--- a/alembic/script/base.py
+++ b/alembic/script/base.py
@@ -782,8 +782,7 @@ class ScriptDirectory:
         message: Optional[str],
         create_date: "datetime.datetime",
     ) -> str:
-        epoch_reference_date = datetime.datetime(1970, 1, 1)
-        epoch = int((create_date - epoch_reference_date).total_seconds())
+        epoch = int(create_date.timestamp())
         slug = "_".join(_slug_re.findall(message or "")).lower()
         if len(slug) > self.truncate_slug_length:
             slug = slug[: self.truncate_slug_length].rsplit("_", 1)[0] + "_"
--- a/alembic/script/base.py
+++ b/alembic/script/base.py
@@ -782,6 +782,8 @@ class ScriptDirectory:
         message: Optional[str],
         create_date: "datetime.datetime",
     ) -> str:
+        epoch_reference_date = datetime.datetime(1970, 1, 1)
+        epoch = int((create_date - epoch_reference_date).total_seconds())
         slug = "_".join(_slug_re.findall(message or "")).lower()
         if len(slug) > self.truncate_slug_length:
             slug = slug[: self.truncate_slug_length].rsplit("_", 1)[0] + "_"
@@ -790,6 +792,7 @@ class ScriptDirectory:
             % {
                 "rev": rev_id,
                 "slug": slug,
+                "epoch": epoch,
                 "year": create_date.year,
                 "month": create_date.month,
                 "day": create_date.day,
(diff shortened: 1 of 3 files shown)


alembic/script/base.py
```diff
@@ -782,6 +782,8 @@ def _rev_path(
         message: Optional[str],
         create_date: "datetime.datetime",
     ) -> str:
+        epoch_reference_date = datetime.datetime(1970, 1, 1)
+        epoch = int((create_date - epoch_reference_date).total_seconds())
```
zzzeek, line 786: I'd rather not hardcode this here.  you can call `create_date.timestamp()`:

```py
>>> import time
>>> import datetime
>>> create_date = datetime.datetime.now()
>>> epoch_reference_date = datetime.datetime(1970, 1, 1)
>>> epoch = int((create_date - epoch_reference_date).total_seconds())
>>> epoch
1651522726
>>> create_date.timestamp()
1651537126.489675

```

zzzeek (changes requested):
we can add this but let's use std library

alembic/script/base.py
zzzeek: I think the latter is using UTC which accounts for the discrepancy

lowercase00: Sure, the reason I chose that path was because going with `timestamp()` would need a `round(..., 0)` and the conversion to `int` to avoid the decimals, but I guess it should work without problems.

Do you think this `int(round(create_date.timestamp()))` could work?

```python
>>> import time
>>> import datetime
>>> create_date = datetime.datetime.now()
>>> epoch_reference_date = datetime.datetime(1970, 1, 1)
>>> epoch = int((create_date - epoch_reference_date).total_seconds())
>>> epoch
1651527473
>>> create_date.timestamp()
1651538273.443539
>>> int(round(create_date.timestamp(),0))
1651538273
>>>
```

zzzeek: yeah just int it, i think that's fine

lowercase00: great, just pushed the new version and fixes tests, everything seems to pass.
thanks a lot for the fast review @zzzeek

CaselIT (approved):
looks ok, but the test fail, could you take a look?

You find the failure in the action checks

lowercase00:
Sure, just found out it is because of the timezone, I'm GMT -3 and tests pass locally, so when running on GMT the epochs differ, I'll take a look on how to fix this.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 27970cd5d4e047cb7aabd161c68b757b2ee3e4c4 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 27970cd5d4e047cb7aabd161c68b757b2ee3e4c4: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/3823

sqla-tester:

thanks for the contribution!

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/3823

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/3823 has been **merged**. Congratulations! :)

PR #1044, opened 2022-05-30
Genarlise env.py to get database URL from .env Fixes: #1043

Zaffer:
### Description
it is better to get your private database url from an .env file. Putting secrets in your ini file is not the best, is it?

### Checklist
This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/templates/generic/env.py
+++ b/alembic/templates/generic/env.py
@@ -30,11 +30,11 @@ target_metadata = None
 # my_important_option = config.get_main_option("my_important_option")
 # ... etc.
 
-postgres_url = os.getenv(
-    "POSTGRES_URL", "postgresql+psycopg2://postgres:postgres@localhost:5432/postgres"
+alembic_url = os.getenv(
+    "ALEMBIC_URL", "postgresql+psycopg2://postgres:postgres@localhost:5432/postgres"
 )
 configuration = config.get_section(config.config_ini_section)
-configuration["sqlalchemy.url"] = postgres_url
+configuration["sqlalchemy.url"] = alembic_url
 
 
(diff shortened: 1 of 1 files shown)


alembic/templates/generic/env.py
```diff
@@ -1,10 +1,15 @@
+import os
 from logging.config import fileConfig
 
+from dotenv import load_dotenv
```
CaselIT, line 4: I don't think the general template should import a 3rd party dependency

alembic/templates/generic/env.py
```diff
@@ -25,6 +30,12 @@
 # my_important_option = config.get_main_option("my_important_option")
 # ... etc.
 
+alembic_url = os.getenv(
+    "ALEMBIC_URL", "postgresql+psycopg2://postgres:postgres@localhost:5432/postgres"
```
CaselIT, line 34: The general template should not assume any url

CaselIT (changes requested):
I think it may make sense to use the env variable `ALEMBIC_URL` for the url and if not found fall-back on the ini file, but the env file should not assume 3rd party packages or a particular db.

zzzeek:
hi -

env.py files are templates that you can modify in any way you see fit, including using your own configuration systems.    the template provided is a generic example only.

alembic/templates/generic/env.py
Zaffer: fair

alembic/templates/generic/env.py
Zaffer: got it

PR #1076, opened 2022-08-11
fix-flake8-errors

Voldemat:
Hi, I use alembic and flake8 in my project. Flake8 raises error on new migrations files, because of no line between alembic and sqlalchemy imports.
Flake8 error: I201 Missing newline between import groups. 'import sqlalchemy' is identified as Third Party and 'from alembic import op' is identified as Third Party.

### DIFF
--- a/alembic/templates/async/script.py.mako
+++ b/alembic/templates/async/script.py.mako
@@ -6,6 +6,7 @@ Create Date: ${create_date}
 
 """
 from alembic import op
+
 import sqlalchemy as sa
 ${imports if imports else ""}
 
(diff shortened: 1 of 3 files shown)


CaselIT:
Hi,

Thanks for the PR, but we are in general -1 on style changes to the default script mako files, since it's impossible to have one configuration that works for every possible linter setups.

Once a project is inited alembic will no longer use the internal scripts but instead uses the one in the alembic folder, so shat you can do is modify the script.py.mako that is in the alembic folder in your install to suit your formatting preference.

Issue #1098, opened 2022-10-12
Postgres Index with opperators dont autogenerate


garyvdm:
**Describe the bug**
I'm using a index with a inet_ops operator.  Autogenerated code does not include the index.

**Expected behavior**
Autogenerated code should include the index.

**To Reproduce**

****env.py***
```py
from logging.config import fileConfig

from sqlalchemy import Column, Index, MetaData, Table, engine_from_config, literal_column
from sqlalchemy.dialects.postgresql import INET
from sqlalchemy import pool

from alembic import context

config = context.config
if config.config_file_name is not None:
    fileConfig(config.config_file_name)

target_metadata = MetaData()

Table(
    "t",
    target_metadata,
    Column("addr", INET),
    Index(
        "ix_1",
        literal_column("addr inet_ops"),
        postgresql_using="GiST",
    ),
)

def run_migrations_offline() -> None:
    url = config.get_main_option("sqlalchemy.url")
    context.configure(
        url=url, target_metadata=target_metadata, 
        literal_binds=True, dialect_opts={"paramstyle": "named"}
    )
    with context.begin_transaction():
        context.run_migrations()

def run_migrations_online() -> None:
    connectable = engine_from_config(
        config.get_section(config.config_ini_section),
        prefix="sqlalchemy.", poolclass=pool.NullPool,
    )
    with connectable.connect() as connection:
        context.configure(connection=connection, target_metadata=target_metadata)
        with context.begin_transaction():
            context.run_migrations()

if context.is_offline_mode():
    run_migrations_offline()
else:
    run_migrations_online()
```

**Error**

```
% alembic revision --autogenerate -m "Add inet ops table/index"
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
INFO  [alembic.autogenerate.compare] Detected added table 't'
/home/gary/dev/alembic-inet-ops-mcve/ve/lib64/python3.10/site-packages/alembic/ddl/postgresql.py:252: UserWarning: autogenerate skipping functional index ix_1; not supported by SQLAlchemy reflection
  util.warn(
  Generating /home/gary/dev/alembic-inet-ops-mcve/alembic/versions/7361d9f63fd9_add_inet_ops_table_index.py ...  done
```

**Versions.**
 - OS: Fedora 36
 - Python: Python 3.10.7
 - Alembic: 1.8.1
 - SQLAlchemy: 1.4.41
[...]

PR #1099, opened 2022-10-12
Postgresql autogenterate: allow for index operators (such as inet_ops)

garyvdm:
This change is so that a Postgresql  index operators (such as inet_ops) gets autogenterated. It does so by broadening the types of index expression allowed from `Column` -> `ColumnClause`.

Fixes: #1098

### Checklist

This pull request is:

- [x] A short code fix
	- Please include: `Fixes: #1098` in the commit message
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/ddl/postgresql.py
+++ b/alembic/ddl/postgresql.py
@@ -248,7 +248,7 @@ class PostgresqlImpl(DefaultImpl):
             for expr in exprs:
                 while isinstance(expr, UnaryExpression):
                     expr = expr.element
-                if not isinstance(expr, Column):
+                if not isinstance(expr, ColumnClause):
                     if sqla_compat.sqla_2:
                         msg = ""
                     else:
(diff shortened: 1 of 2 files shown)


CaselIT:
Hi,

Thanks for starting the implementation on this.
I'm not sure if it's this easy the resolution, since in any case sqlalchemy 1.4 will not reflect these indexes, so the current logic would need to continue to stay in place.

The autogenerate compare would also probably need to take these into account

garyvdm:
Ok, I'm going try improve the change to address your comment.

garyvdm:
So I have done a lot of reading code, and more experimentation. 

> I'm not sure if it's this easy the resolution, since in any case sqlalchemy 1.4 will not reflect these indexes, so the current logic would need to continue to stay in place.

Ok, I finally understand what you mean here. 2.0.0b2 also doesn't reflect all information needed. I'm going to log a bug for that, and try see if I can fix that. 

In the mean time, I'm going to close this PR.

CaselIT:
Oh understood what you mean by index with operations. I though you meant an index with expressions like `lower(col), upper(other_col)`. Sorry.

Yes, I'm not sure if this information is reflected at the moment, but I don't think so

PR #1140, opened 2022-12-08
Bug correction that impeded the "Naming Columns Distinctly from Attribute Names" feature of sqlachemy

ceprio:
Found a code misshapps that would generate a "KeyError: 'Key Name'" when alembic is autogenerating the difference between existing metadata and reflected metadata. The problem arises when the existing metadata contains something like:

```
class Links(Base):
    __tablename__ = "Links"
    id = Column('Id', Integer, primary_key=True)
    webpage = Column('Web page', String(50), default=None)
```

P.S. Notice the webpage definition that is named differently than the actual table column name.

### Description
The changed code simply uses the right column dictionnary that is created with the database naming instead of the users metadata attribute name. (i.e. using conn_col_names instead of conn_table.c).

### Checklist
-->

This pull request is:

- [x ] A short code fix
	- please include tests.

I do not know what kind of test you want for this but your standard test suite should be sufficient to validate this change. I would howerver encourage you to add a different column attribute than the database column in the database definition of your test suite to insure this is supported in the future. See my `Link` class as an example.

Thank you,

### DIFF
--- a/alembic/autogenerate/compare.py
+++ b/alembic/autogenerate/compare.py
@@ -371,7 +371,7 @@ def _compare_columns(
 
     for colname in metadata_col_names.intersection(conn_col_names):
         metadata_col = metadata_cols_by_name[colname]
-        conn_col = conn_table.c[colname]
+        conn_col = conn_col_names[colname]
         if not autogen_context.run_object_filters(
             metadata_col, colname, "column", False, conn_col
         ):


CaselIT:
Hi,

Sorry, but this need to first have an issue opened.
The provided example does not illustrate any bug.

If I try making your example runnable there is no reported issue, so I don't know what this PR is trying to fix.
Please open an issue as instructed by the PR template to explain the issue, Thanks

```py
import sqlalchemy as sa
import sqlalchemy.orm as orm

class Base(orm.DeclarativeBase):
    pass

class Foo(Base):
    __tablename__ = "Links"
    id = sa.Column('Id', sa.Integer, primary_key=True)
    webpage = sa.Column('Web page', sa.String(50), default=None)

e = sa.create_engine('sqlite://')
Base.metadata.create_all(e)

meta = sa.MetaData()
ref = sa.Table('Links', meta)
sa.inspect(e).reflect_table(ref, None)
print(ref.c['Id'], ref.c['Web page'])
```

PR #1172, opened 2023-02-11
Include missing part of env.py

Dreamsorcerer:
This part is missing from the example, and if a user started with the async template, then they would have an `asyncio.run()` call in here, which breaks the script.

### DIFF
--- a/docs/build/cookbook.rst
+++ b/docs/build/cookbook.rst
@@ -1562,6 +1562,12 @@ together, and ``env.py`` as follows works::
         else:
             do_run_migrations(connectable)
 
+
+    if context.is_offline_mode():
+        run_migrations_offline()
+    else:
+        run_migrations_online()
+
 Above, using an asyncio database URL in ``alembic.ini`` one can run
 commands such as ``alembic upgrade`` from the command line.  Programmatically,
 the same ``env.py`` file can be invoked using asyncio as::


sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision fa2c55319d09fab4e1435433fa2d7e5fca5dc820 of this pull request into gerrit so we can run tests and reviews and stuff

Dreamsorcerer:
@CaselIT This is part of the "Programmatic API use (connection sharing) With Asyncio" example in the documentation. I don't think the templates need changing, as they are for a different use case.

Dreamsorcerer:
Unless, of course, you want to make this behaviour available by default..

CaselIT:
yes I did not mean that, I mean like how the engine is created.

Personally I think it makes more sense to use use the same structure as in the sync version, so illustrating only the changes to `run_migrations_online`

CaselIT:
@Dreamsorcerer I've updated at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4443 if you want to take a look

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4443 has been **merged**. Congratulations! :)

sqla-tester:

uh, I did not want to actually merge, pushed by mistake

Mike feel free to change as you prefer

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4443

sqla-tester (commented):

I've replied

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4443

* /COMMIT_MSG (line 9): so, currently the cookbook does:
do these changes to your existing template to use it with asyncio: https://alembic.sqlalchemy.org/en/latest/cookbook.html#using-asyncio-with-alembic

then later there is "if you want to use alembic programmatically, you can do these changes to the above template to make it work": https://alembic.sqlalchemy.org/en/latest/cookbook.html#programmatic-api-use-connection-sharing-with-asyncio

the issue is that with the changes in that section something the section that launched the two functions needed to change.
* docs/build/cookbook.rst (line 1523): because it's the same as the template in https://alembic.sqlalchemy.org/en/latest/cookbook.html#using-asyncio-with-alembic so there is no need to list it here, since it's not a change from that template

sqla-tester:

nope nothing on this end :)


cant believe how many changes / questions/ discsusions we have to field per day this is too much

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4443

sqla-tester:

indeeed. I've the email that's asking for a day off :\

I'll look into doing these changes

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4443

sqla-tester (commented):

added a new follow up change with your suggestions

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4443

* docs/build/cookbook.rst (line 1523): we can continue in I131dfb8dabf10457954f5216b1abadc976aebd38

PR #1173, opened 2023-02-12
Update annotation to allow PathLike

Dreamsorcerer:
The value is only used in os.path.abspath(), so it works fine with Path.

### DIFF
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -10,6 +10,7 @@ from typing import Dict
 from typing import Optional
 from typing import overload
 from typing import TextIO
+from typing import Union
 
 from . import __version__
 from . import command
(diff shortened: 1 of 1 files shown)


alembic/config.py
```diff
@@ -123,7 +124,7 @@ def __init__(
 
     """
 
-    config_file_name: Optional[str] = None
+    config_file_name: Union[str, os.PathLike[str], None] = None
```
Dreamsorcerer, line 127: This annotation is actually redundant, you could remove this line completely.

sqla-tester:
New Gerrit review created for change 7cf9606602bb5692ddeb3551500be0dd956bd5d9: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4444

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4444 has been **merged**. Congratulations! :)

Issue #1184, opened 2023-02-21
literal_column() in ExcludeConstraint are autogenerated as column(), failing in the migration


jankatins:
**Describe the bug**

If you add a literal_column in a ExcludeConstraint (e.g. a `literal_column("int8range(col_from, col_to)")`), this is rendered as `column("int8range(col_from, col_to)")` in the autogenerated migration and the migration will fail as such a column does not exist in the table.

**Expected behavior**

literal_columns are passed through in the rendered migration

**To Reproduce**
```py
from sqlalchemy.dialects.postgresql import JSONB, ExcludeConstraint
from sqlalchemy.sql.expression import column, literal_column
from sqlalchemy import Column, Table
from sqlalchemy.orm import declarative_base
from sqlalchemy.sql.schema import MetaData
from sqlalchemy.sql.sqltypes import *

Base2 = declarative_base()
whatever2 = Table(
    "whatever2",
    Base2.metadata,
    Column("id", BigInteger(), nullable=False, primary_key=True),
    ExcludeConstraint(
        ("id", "="),
        (literal_column("id + 2"), "="),
        name=f"whatever_id_int8range_excl",
    ),
)
```

**Error**

The migration code after `alembic revision --autogenerate -m "test2"`looks like this:

```py
def upgrade() -> None:
    # ### commands auto generated by Alembic - please adjust! ###
    op.create_table('whatever2',
    sa.Column('id', sa.BigInteger(), nullable=False),
    postgresql.ExcludeConstraint((sa.column('id'), '='), (sa.column('id + 2'), '='), using='gist', name='whatever_id_int8range_excl'),
    sa.PrimaryKeyConstraint('id')
    )
```

running this migration via `alembic upgrade head` results in:

```
psycopg2.errors.UndefinedColumn: column "id + 2" named in key does not exist


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/jankatins/Library/Caches/pypoetry/virtualenvs/mds-elt-service-JeB-l5pZ-py3.10/bin/alembic", line 8, in <module>
    sys.exit(main())
  File "/Users/jankatins/Library/Caches/pypoetry/virtualenvs/mds-elt-service-JeB-l5pZ-py3.10/lib/python3.10/site-packages/alembic/config.py", line 591, in main
    CommandLine(prog=prog).main(argv=argv)
[...]

PR #1185, opened 2023-02-21
Handle literal_column in ExcludeConstraint differently

jankatins:
render `literal_columns()` as `sa.literal_columns()` to allow for expressions in ExcludeConstraint.

Fixes: https://github.com/sqlalchemy/alembic/issues/1184

### Description

Before the change, a literal_column in an ExcludeConstraint ended up as a `column("...")`which in turn meant that the actual migration errored with

```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedColumn) column "id + 2" named in key does not exist

[SQL:
CREATE TABLE whatever2 (
        id BIGSERIAL NOT NULL,
        PRIMARY KEY (id),
        CONSTRAINT whatever_id_int8range_excl EXCLUDE USING gist (id WITH =, "id + 2" WITH =)
)
```

Note: the column name is interpreted as a quoted column named "id + 2".

After the fix, a literal_column is passed through, ending up in the correct SQL, which can be run on PG.

### Checklist

This pull request is:

- [x] A short code fix

### DIFF
--- a/alembic/ddl/postgresql.py
+++ b/alembic/ddl/postgresql.py
@@ -22,6 +22,7 @@ from sqlalchemy.dialects.postgresql import ExcludeConstraint
 from sqlalchemy.dialects.postgresql import INTEGER
 from sqlalchemy.schema import CreateIndex
 from sqlalchemy.sql.elements import ColumnClause
+from sqlalchemy.sql.elements import TextClause
 from sqlalchemy.types import NULLTYPE
 
 from .base import alter_column
@@ -667,11 +668,11 @@ def _exclude_constraint(
 
 
 def _render_potential_column(
-    value: Union[ColumnClause, Column], autogen_context: AutogenContext
+    value: Union[ColumnClause, Column, TextClause], autogen_context: AutogenContext
 ) -> str:
     if isinstance(value, ColumnClause):
         if value.is_literal:
-            # Support stuff like literal_column("int8range(from, to)") in ExcludeConstraint
+            # like literal_column("int8range(from, to)") in ExcludeConstraint
             template = "%(prefix)sliteral_column(%(name)r)"
         else:
             template = "%(prefix)scolumn(%(name)r)"
@@ -680,7 +681,12 @@ def _render_potential_column(
             "prefix": render._sqlalchemy_autogenerate_prefix(autogen_context),
             "name": value.name,
         }
-
+    elif isinstance(value, TextClause):
+        template = "%(prefix)stext(%(text)r)"
+        return template % {
+            "prefix": render._sqlalchemy_autogenerate_prefix(autogen_context),
+            "text": value.text,
+        }
     else:
         return render._render_potential_expr(
             value, autogen_context, wrap_in_text=False
--- a/alembic/ddl/postgresql.py
+++ b/alembic/ddl/postgresql.py
@@ -670,7 +670,11 @@ def _render_potential_column(
     value: Union[ColumnClause, Column], autogen_context: AutogenContext
 ) -> str:
     if isinstance(value, ColumnClause):
-        template = "%(prefix)scolumn(%(name)r)"
+        if value.is_literal:
+            # Support stuff like literal_column("int8range(from, to)") in ExcludeConstraint
+            template = "%(prefix)sliteral_column(%(name)r)"
+        else:
+            template = "%(prefix)scolumn(%(name)r)"
 
         return template % {
             "prefix": render._sqlalchemy_autogenerate_prefix(autogen_context),
--- a/tests/test_postgresql.py
+++ b/tests/test_postgresql.py
@@ -1182,6 +1182,36 @@ class PostgresqlAutogenRenderTest(TestBase):
             "name='TExclID'))",
         )
 
+    @config.requirements.sqlalchemy_2
+    def test_inline_exclude_constraint_text(self):
+        # Actually, this only works on sqlalchemy 2.x >= 2023-03-03 due to a bug in sqlalchemy
+        # See https://github.com/sqlalchemy/alembic/issues/1184
+        #     and https://github.com/sqlalchemy/sqlalchemy/issues/9401
+
+        from sqlalchemy.dialects.postgresql import ExcludeConstraint
+
+        autogen_context = self.autogen_context
+
+        m = MetaData()
+        t = Table(
+            "TTable", m, Column("id", String()),
+            ExcludeConstraint(
+                (text("id + 2"), "="), name="TExclID", using="gist"
+            ),
+        )
+
+        op_obj = ops.CreateTableOp.from_table(t)
+
+        eq_ignore_whitespace(
+            autogenerate.render_op_text(autogen_context, op_obj),
+            "op.create_table('TTable',sa.Column('id', sa.String(), "
+            "nullable=True),"
+            "postgresql.ExcludeConstraint((sa.text('id + 2'), '='), "
+            "using='gist', "
+            "name='TExclID'))",
+        )
+
+
     def test_json_type(self):
         eq_ignore_whitespace(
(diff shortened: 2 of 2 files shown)


CaselIT (commented):
The change looks ok. If you could also support text it would be great.
I think you can wrap it into `text(...)`

Thanks for the report and the PR!

CaselIT:
that's strange. What is the exclude constraint that returns that error?

Also maybe I was not clear above. I meant:
- keep the implementation originally proposed in this PR for literal_columns
- additionally support text. it may already work, but there are no test in alembic, so it should be checked.

let me know if you want me to take over and finish it

jankatins:
Pushed the text change which produced the above error as https://github.com/sqlalchemy/alembic/pull/1185/commits/90d4884effd5f5ae1dab25a913910e042eadd662

CaselIT:
Ok, I will try to tale a look tomorrow

CaselIT:
ok this is another bug in sqlalchemy. ExcludeConstraint can't be copied or if it can it's all kind of broken

```py
import sqlalchemy as sa
from sqlalchemy.dialects.postgresql import ExcludeConstraint
from sqlalchemy.schema import AddConstraint

t = sa.Table(
    "x",
    sa.MetaData(),
    sa.Column("a"),
    sa.Column("b"),
    c := ExcludeConstraint((sa.text("a+1"), "a")),
    c2:=ExcludeConstraint((sa.column("a") + sa.column("b"), "a"))
)


print(AddConstraint(c))
try:
    cc = c._copy()
    cc._set_parent(t)
    print(AddConstraint(cc))
except Exception as e:
    print(e)
print(AddConstraint(c2))
cc2 = c2._copy()
cc2._set_parent(t)
print(AddConstraint(cc2))
```
prints
```
ALTER TABLE x ADD EXCLUDE USING gist (a+1 WITH a)
not enough values to unpack (expected 2, got 0)
ALTER TABLE x ADD EXCLUDE USING gist (a + b WITH a)
ALTER TABLE x ADD EXCLUDE USING gist (a WITH a)
```

@zzzeek are the two form above supported?

zzzeek:
> ok this is another bug in sqlalchemy. ExcludeConstraint can't be copied or if it can it's all kind of broken
> 
> ```python
> import sqlalchemy as sa
> from sqlalchemy.dialects.postgresql import ExcludeConstraint
> from sqlalchemy.schema import AddConstraint
> 
> t = sa.Table(
>     "x",
>     sa.MetaData(),
>     sa.Column("a"),
>     sa.Column("b"),
>     c := ExcludeConstraint((sa.text("a+1"), "a")),
>     c2:=ExcludeConstraint((sa.column("a") + sa.column("b"), "a"))
> )
> 
> 
> print(AddConstraint(c))
> try:
>     cc = c._copy()
>     cc._set_parent(t)
>     print(AddConstraint(cc))
> except Exception as e:
>     print(e)
> print(AddConstraint(c2))
> cc2 = c2._copy()
> cc2._set_parent(t)
> print(AddConstraint(cc2))
> ```
> 
> prints
> 
> ```
> ALTER TABLE x ADD EXCLUDE USING gist (a+1 WITH a)
> not enough values to unpack (expected 2, got 0)
> ALTER TABLE x ADD EXCLUDE USING gist (a + b WITH a)
> ALTER TABLE x ADD EXCLUDE USING gist (a WITH a)
> ```
> 
> @zzzeek are the two form above supported?

copying of ExcludeConstraint is supported within the context of `to_metadata()`.   I dont know that its `_copy()` method does not require additional context to be passed in order to function correctly.

CaselIT:
_copy() uses columns that don't work at all for text and report only a single column for expressions.

If both are supported, copy needs to be updated

zzzeek:
> _copy() uses columns that don't work at all for text (text element don't end up in columns) and report only a single column for expressions.
> 
> If both are supported, copy needs to be updated

the text() and column() elements I see in your example both dont need to be copied in this case, they can be used in place.

alembic/ddl/postgresql.py
```diff
@@ -650,10 +651,16 @@ def _exclude_constraint(
 
 
 def _render_potential_column(
-    value: Union[ColumnClause, Column], autogen_context: AutogenContext
+    value: Union[ColumnClause, Column, TextClause], autogen_context: AutogenContext
 ) -> str:
-    if isinstance(value, ColumnClause):
-        template = "%(prefix)scolumn(%(name)r)"
+    if isinstance(value, (ColumnClause, TextClause)):
```
zzzeek, line 656: whatever is being fixed here seems like it should apply to Index also

CaselIT: I can look into index to see how it behaves. I think there are tests of expressions and text in alembic with indexes

CaselIT:
the issue in sqlalchemy are fixed in https://github.com/sqlalchemy/sqlalchemy/issues/9401

CaselIT:
sqlalchemy issue was fixted. let's see if the actions are happier now

jankatins:
I see that [3.10-sqlamain-ubuntu-latest](https://github.com/sqlalchemy/alembic/actions/runs/4287272593/jobs/7538008180#logs) works, but the old sqla versions won't. Seems I have to add some pytest skips for these versions?

jankatins:
Skipped the text tests on sqla<2.x (seems gh actions only run on 2.x head, so that should make these work. Didn't see a way to test for specific version of sqla :-()

Also fixed the mypy problem, hopefully.

-> Action run needs approval @CaselIT

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 68360ce9aa746a85407c5e1e04b8021123d98504 of this pull request into gerrit so we can run tests and reviews and stuff

alembic/ddl/postgresql.py
CaselIT: ok, index just renders everything as create_index(..., [sa.text(), sa.text()]) since it never renders it inline it does not have an issue with it. In particular it will convert even function into `text()`.

exclude constraint cannot really do that since it can be rendered inline and until we fixed it it did not handle text at all.

so I think this could be a separate function, but maybe the logic could be moved to rendered.py?

sqla-tester:
**Jan Katins** has proposed a fix for this issue in the **main** branch:

**Handle literal_column in ExcludeConstraint differently** https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4493

tests/test_postgresql.py
```diff
@@ -1156,6 +1157,61 @@ def test_inline_exclude_constraint_case_sensitive(self):
             "name='TExclX'))",
         )
 
+    def test_inline_exclude_constraint_literal_column(self):
+        # Issue: https://github.com/sqlalchemy/alembic/issues/1184
```
sqla-tester, line 1161: **mike bayer** (zzzeek) wrote:

it would be preferred to label this like:

   
    """fix for #1184"""

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4493

sqla-tester (commented):

needs a changelog file, add credit for the PR author as usual...

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4493

CaselIT:
1. Could not use literal_column() in exclude contraints
2. You can do that now. Also since sqlalchemy 2.0.5 you can also use text()
3. Any should work. To use test() you need 2.0.5
4. Not worse than before

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4493 has been **merged**. Congratulations! :)

jankatins:
@CaselIT @zzzeek Thanks a lot for fixing this up and getting it merged!

CaselIT:
thanks for the effort!

PR #1218, opened 2023-04-09
tooling: write_pyi.py: filter usage of raw-strings (rstrings)

jayaddison:
### Description
While reading the diff between [`rel_1_10_2...rel_1_10_3`](https://github.com/sqlalchemy/alembic/compare/rel_1_10_2...rel_1_10_3), the introduction of r-strings drew my attention, and that resulted in some [discussion on the relevant commit](https://github.com/sqlalchemy/alembic/commit/bc0c305b7c2cc0401e250fcd6a725aacecdd6e33).

This changeset filters the production of r-strings during stub-generation to cases where docstrings contain escape (backslash, `\`) characters.

I'll admit that I didn't realize until today that these stubs are primarily for typechecking.  Since that's the case, I have doubts about whether the change is worthwhile (consistency and simplicity -- using r-strings for all docstrings in the stubs -- even if it's redundant, seems fine to me).

cc @CaselIT 

### Checklist
This pull request is:

- [x] A short code fix
	- Relates to discussion at https://github.com/sqlalchemy/alembic/commit/bc0c305b7c2cc0401e250fcd6a725aacecdd6e33

### DIFF
--- a/tools/write_pyi.py
+++ b/tools/write_pyi.py
@@ -218,7 +218,8 @@ def _generate_stub_for_meth(
 
     fn_doc = base_method.__doc__ if base_method else fn.__doc__
     has_docs = gen_docs and fn_doc is not None
-    docs = 'r"""' + f"{fn_doc}" + '"""' if has_docs else ""
+    string_prefix = "r" if chr(92) in fn_doc else ""
+    docs = f'{string_prefix}"""' + f"{fn_doc}" + '"""' if has_docs else ""
 
     func_text = textwrap.dedent(
         f"""


CaselIT:
Thanks, I'll take a look soon

tools/write_pyi.py
```diff
@@ -218,7 +218,8 @@ def _formatvalue(value):
 
     fn_doc = base_method.__doc__ if base_method else fn.__doc__
     has_docs = gen_docs and fn_doc is not None
-    docs = 'r"""' + f"{fn_doc}" + '"""' if has_docs else ""
+    string_prefix = "r" if chr(92) in fn_doc else ""
```
CaselIT, line 221: this is probably a better regexp `r'\\[^\\]'` should match single unescaped `\` only

jayaddison: I'm not sure about that - it could miss some cases [like this one](https://github.com/sqlalchemy/alembic/blob/1ae44556907f14f92874f05d05242cb57bb0f855/alembic/operations/ops.py#L2346)?

jayaddison:
Another (not yet handled in this branch) edge case that could use testing before any changes are applied here: a source docstring declared using r-string literal format, with contents that begin with a double-quote (`"`) character.

tools/write_pyi.py
CaselIT: indeed that does not seem to work on that string.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 352ab8829525435f78838687e12156a3fa3a3d86 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 352ab8829525435f78838687e12156a3fa3a3d86: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4561

sqla-tester:

I've not been following the discussion what this is about.

can we summarize succinctly what the problem is, and how this solves it?   We can also add a changelog file indicating that.    

for example I dont know what the problem with "r" strings is

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4561

sqla-tester:
**James Addison** (jayaddison) wrote:

(discussion-starter here)

I don't have any significant technical reasons to suggest for doing this.

It's not causing any problems, and the files where the r-strings appear are .pyi stubs (which as I understand it, aren't involved in runtime control-flow).

While reviewing the plaintext diff from 1.10.2 to 1.10.3, my sense was that it was a potential code smell to begin using a different literal string representation for docstrings, particularly in cases that don't require character escaping -- but it's also harmless, afaik.

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4561

sqla-tester (commented):

OK I guess acutal code change is one line, looks like possible typo in setup.cfg?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4561

* setup.cfg (line 91): is this a typo?

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4561 has been **merged**. Congratulations! :)

jayaddison:
Thank you @CaselIT!

PR #1219, opened 2023-04-09
ops.py: cleanup / consistency: remove redundant docstring newline markers

jayaddison:
### Description
Potential (nitpicky) docstring cleanup discovered while preparing #1218.

### Checklist

This pull request is:

### DIFF
--- a/alembic/op.pyi
+++ b/alembic/op.pyi
@@ -969,10 +969,10 @@ def execute(
             column('name', String)
         )
         op.execute(
-            account.update().\\
-                where(account.c.name==op.inline_literal('account 1')).\\
-                values({'name':op.inline_literal('account 2')})
-                )
+            account.update().
+            where(account.c.name==op.inline_literal('account 1')).
+            values({'name':op.inline_literal('account 2')})
+        )
 
     Above, we made use of the SQLAlchemy
     :func:`sqlalchemy.sql.expression.table` and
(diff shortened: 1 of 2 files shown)


jayaddison:
The docstring here provides a sample case that could be useful for #1218 (in-progress), so I'm going to move this pull request into a draft status too.

CaselIT:
Hi,

if we change that part we should probably format the text using black.

jayaddison:
@CaselIT that's OK with me :+1: 

If you have suggestions for additional ways to provide test coverage for the script, let me know and I'll take a look.

I'm still learning about `.pyi` stub files (not to mention `.pyx` that I see elsewhere in the codebase) and this could be a good way to learn more.

CaselIT:
> not to mention .pyx that I see elsewhere in the codebase

these are [cython](https://cython.org/) files, and should only be on sqlalchemy side. there is no performance sensitive code on alembic :)

jayaddison:
Self-replying:
> If you have suggestions for additional ways to provide test coverage for the script, let me know and I'll take a look.

Thinking about this more over the past day or so: I suppose the fact that the `write_pyi.py` script [internally runs `zimports` and `black`](https://github.com/sqlalchemy/alembic/blob/d3f869bb3159d18669fb6f00cbe514307118eea7/tools/write_pyi.py#L137-L149) on the output is a kind of self-contained test coverage.

I'm also thinking about going through some of the code samples within docstrings in the codebase to apply `black` code formatting to those, for consistency in the documentation.

CaselIT:
by multiline strings you mean a single text that spans multiple python line, but that does not include `\n` or instead something that's enclosed in `'''` to have multiple `\n` in the text?

CaselIT:
It's maybe simple in these case to format with black then re-convert `"""` to `'''` so we avoid issues with the doc strings that sometimes are `'''`

PR #1236, opened 2023-05-08
alembic/ddl/sqlite.py: enhance the judgment condition to fix the problem of AttributeError: 'NoneType' object has no attribute 'persisted' in the requires_recreate_in_batch function.

hustliyilin:
Signed-off-by: YiLin.Li <YiLin.Li@linux.alibaba.com>

### DIFF
--- a/alembic/ddl/sqlite.py
+++ b/alembic/ddl/sqlite.py
@@ -60,7 +60,7 @@ class SQLiteImpl(DefaultImpl):
                     return True
                 elif (
                     isinstance(col.server_default, util.sqla_compat.Computed)
-                    and col.server_default.persisted
+                    and col.server_default and col.server_default.persisted
                 ):
                     return True
             elif op[0] not in ("create_index", "drop_index"):


alembic/ddl/sqlite.py
```diff
@@ -60,7 +60,7 @@ def requires_recreate_in_batch(
                     return True
                 elif (
                     isinstance(col.server_default, util.sqla_compat.Computed)
-                    and col.server_default.persisted
+                    and col.server_default and col.server_default.persisted
```
CaselIT, line 63: Ok, this is because of the mock we do in sqla_compat.

Thanks for reporting, but the fix should be something else.

CaselIT (changes requested):
I'll create an issue

alembic/ddl/sqlite.py
hustliyilin: OK. If you fix this issue, please let me know. Thanks.

hustliyilin:
OK，Thanks.

Issue #1273, opened 2023-07-14
Use an explicit encoding when reading the configuration file


kkirsche:
**Describe the bug**

When the python option `PYTHONWARNDEFAULTENCODING` is set as described on https://docs.python.org/3/library/io.html#opt-in-encodingwarning

```
/Users/kkirsche/Library/Caches/pypoetry/virtualenvs/application-7pVymVLv-py3.11/lib/python3.11/site-packages/alembic/config.py:204: EncodingWarning: 'encoding' argument not specified
  file_config.read([self.config_file_name])
```

**Expected behavior**
Alembic will explicitly specify the encoding to use when reading the configuration file. Per https://docs.python.org/3/library/io.html#opt-in-encodingwarning, the recommendation would be to use `"utf-8"`.

**To Reproduce**
Run an alembic migration with the environment variable`PYTHONWARNDEFAULTENCODING` set to any value. In my testing, I am using `1` as my value.

**Error**

```
/Users/kkirsche/Library/Caches/pypoetry/virtualenvs/application-7pVymVLv-py3.11/lib/python3.11/site-packages/alembic/config.py:204: EncodingWarning: 'encoding' argument not specified
  file_config.read([self.config_file_name])
```

**Versions.**
 - OS: `macOS 13.4.1 (c) (22F770820d)`
 - Python: `Python 3.11.4`
 - Alembic: `1.11.1`
 - SQLAlchemy: `2.0.18[mysql, mypy]`
 - Database: `mariadb from 11.0.2-MariaDB, client 15.2 for osx10.18 (arm64) using  EditLine wrapper`
 - DBAPI: `mysqlclient 2.2.0`

**Additional context**
The `read` function passes the encoding to https://docs.python.org/3/library/io.html#io.text_encoding

For me, this returns `"locale"` which may make sense to use if you are not comfortable with `"utf-8"`. `"locale"` which, as I understand it, uses https://docs.python.org/3/library/locale.html to detect the encoding settings of the caller's machine.

Thank you for your consideration

**Have a nice day!**

zzzeek:
that's fine, can you please send a pull request?   thanks

kkirsche:
Course, sent 🙂 

Thank you for your time and consideration.

sqla-tester:
**Kevin Kirsche** has proposed a fix for this issue in the **main** branch:

[...]

PR #1274, opened 2023-07-14
Specify `"locale"` encoding when reading configuration file

kkirsche:
### Description
This merge request explicitly sets the encoding of the configuration file being read to "utf-8" as recommended by https://docs.python.org/3/library/io.html#opt-in-encodingwarning

Fixes: https://github.com/sqlalchemy/alembic/issues/1273

### Checklist

This pull request is:

- [X] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -201,9 +201,7 @@ class Config:
         self.config_args["here"] = here
         file_config = ConfigParser(self.config_args)
         if self.config_file_name:
-            # python >=3.10 supports "locale" instead of locale.getencoding()
-            # https://docs.python.org/3/whatsnew/3.10.html#optional-encodingwarning-and-encoding-locale-option
-            file_config.read([self.config_file_name], getencoding())
+            compat.read_config_parser(file_config, [self.config_file_name])
         else:
             file_config.add_section(self.config_ini_section)
         return file_config
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -200,7 +200,7 @@ class Config:
         self.config_args["here"] = here
         file_config = ConfigParser(self.config_args)
         if self.config_file_name:
-            file_config.read([self.config_file_name], "utf-8")
+            file_config.read([self.config_file_name], "locale")
         else:
             file_config.add_section(self.config_ini_section)
         return file_config
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -6,6 +6,7 @@ from configparser import ConfigParser
 import inspect
 import os
 import sys
+from locale import getencoding
 from typing import Any
 from typing import cast
 from typing import Dict
@@ -200,7 +201,9 @@ class Config:
         self.config_args["here"] = here
         file_config = ConfigParser(self.config_args)
         if self.config_file_name:
-            file_config.read([self.config_file_name], "locale")
+            # python >=3.10 supports "locale" instead of locale.getencoding()
+            # https://docs.python.org/3/whatsnew/3.10.html#optional-encodingwarning-and-encoding-locale-option
+            file_config.read([self.config_file_name], getencoding())
         else:
             file_config.add_section(self.config_ini_section)
         return file_config
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -6,7 +6,6 @@ from configparser import ConfigParser
 import inspect
 import os
 import sys
-from locale import getencoding
 from typing import Any
 from typing import cast
 from typing import Dict
--- a/alembic/config.py
+++ b/alembic/config.py
@@ -200,7 +200,7 @@ class Config:
         self.config_args["here"] = here
         file_config = ConfigParser(self.config_args)
         if self.config_file_name:
-            file_config.read([self.config_file_name])
+            file_config.read([self.config_file_name], "utf-8")
         else:
             file_config.add_section(self.config_ini_section)
         return file_config
--- a/alembic/util/compat.py
+++ b/alembic/util/compat.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 import io
 import os
 import sys
-import typing
 from configparser import ConfigParser
+import typing
 from typing import Sequence
 from typing import Union
 
--- a/alembic/util/compat.py
+++ b/alembic/util/compat.py
@@ -4,7 +4,9 @@ import io
 import os
 import sys
 import typing
+from configparser import ConfigParser
 from typing import Sequence
+from typing import Union
 
 from sqlalchemy.util import inspect_getfullargspec  # noqa
 from sqlalchemy.util.compat import inspect_formatargspec  # noqa
@@ -60,7 +62,11 @@ def formatannotation_fwdref(annotation, base_module=None):
         return repr(annotation).replace("~", "")
     return repr(annotation).replace("~", "")
 
-def read_config_parser(file_config, file_argument):
+
+def read_config_parser(
+    file_config: ConfigParser,
+    file_argument: Sequence[Union[str, os.PathLike[str]]],
+) -> list[str]:
     if py310:
         return file_config.read(file_argument, encoding="locale")
     else:
--- a/alembic/util/compat.py
+++ b/alembic/util/compat.py
@@ -12,6 +12,7 @@ from sqlalchemy.util.compat import inspect_formatargspec  # noqa
 is_posix = os.name == "posix"
 
 py311 = sys.version_info >= (3, 11)
+py310 = sys.version_info >= (3, 10)
 py39 = sys.version_info >= (3, 9)
 py38 = sys.version_info >= (3, 8)
 
@@ -58,3 +59,9 @@ def formatannotation_fwdref(annotation, base_module=None):
     elif isinstance(annotation, typing.TypeVar):
         return repr(annotation).replace("~", "")
     return repr(annotation).replace("~", "")
+
+def read_config_parser(file_config, file_argument):
+    if py310:
+        return file_config.read(file_argument, encoding="locale")
+    else:
+        return file_config.read(file_argument)


alembic/config.py
```diff
@@ -200,7 +200,7 @@ def file_config(self):
         self.config_args["here"] = here
         file_config = ConfigParser(self.config_args)
         if self.config_file_name:
-            file_config.read([self.config_file_name])
+            file_config.read([self.config_file_name], "utf-8")
```
CaselIT, line 203: this should be ` locale.getencoding()`

setting this to utf-8 is a possible breaking change if people use strange encodings.

Now I don't expect this to be the case, but either leave the default (so what's done now) or we should use the default format for the locale

CaselIT (changes requested):

alembic/config.py
kkirsche: I don't believe that's correct. If you would like locale, my understanding is that this should be "locale" which would allow read to use the proper locale option. Yes, that resolves to that ultimately, but it ultimately requires more maintenance as you will have to keep track of how to resolve the locale instead of allowing the standard library to do so for you.

kkirsche: Maybe I misunderstood https://docs.python.org/3/library/io.html#io.text_encoding though.

But my thought process is that `.read` is implemented as:

```
    def read(self, filenames, encoding=None):
        """Read and parse a filename or an iterable of filenames.

        Files that cannot be opened are silently ignored; this is
        designed so that you can specify an iterable of potential
        configuration file locations (e.g. current directory, user's
        home directory, systemwide directory), and all existing
        configuration files in the iterable will be read.  A single
        filename may also be given.

        Return list of successfully read files.
        """
        if isinstance(filenames, (str, bytes, os.PathLike)):
            filenames = [filenames]
        encoding = io.text_encoding(encoding)
        read_ok = []
        for filename in filenames:
            try:
                with open(filename, encoding=encoding) as fp:
                    self._read(fp, filename)
            except OSError:
                continue
            if isinstance(filename, os.PathLike):
                filename = os.fspath(filename)
            read_ok.append(filename)
        return read_ok
```

So `io.text_encoding` would return `locale` if passed in or `"utf-8"` / `"locale"` if `None` (currently the case). Since locale is a valid option as seen here, this is what seems to make sense for your ask.

This then goes to `open` which is discussed in https://docs.python.org/3/library/functions.html#open

> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) returns), but any [text encoding](https://docs.python.org/3/glossary.html#term-text-encoding) supported by Python can be used. See the [codecs](https://docs.python.org/3/library/codecs.html#module-codecs) module for the list of supported encodings.

So while ultimately it's the same code, this seems like it's adding unnecessary maintenance overhead should this change in a future python version.

CaselIT: Ok, that works

kkirsche: That's set in the latest commit. If there are any issues though, by all means please let me know and I'll address that as soon as possible for you 👍

Thanks for your time reviewing and your consideration.

CaselIT (approved):
Thanks, let's see run the ci

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 258baabcc4fe5ddf5f983120a3aadb74db8d5c1c of this pull request into gerrit so we can run tests and reviews and stuff

CaselIT:
@zzzeek the bot broke :(

same also on https://github.com/sqlalchemy/sqlalchemy/pull/10106

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 258baabcc4fe5ddf5f983120a3aadb74db8d5c1c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 258baabcc4fe5ddf5f983120a3aadb74db8d5c1c: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4786

CaselIT:
It's alive! 🤖

kkirsche:
It looks like CI failed on 3.9 if I read the Jenkins output right (I don't use Jenkins often so I may have misread it) but worked on later versions. As a result, it seems like support for `"locale"` may have been introduced in newer versions of Python only (seems strange, but I'll have to check the docs, I guess). 

As a result, I'll update the code as originally suggested by @CaselIT and then add a comment indicating why that solution was used instead of `"locale"`, which will be usable whenever you choose to deprecate 3.9 (if when it becomes EOL per https://devguide.python.org/versions/ that would be 2025-Oct)

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision c0d18a0c6126c5561f29146813bf751a5ee2496a of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset c0d18a0c6126c5561f29146813bf751a5ee2496a added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4786

CaselIT:
turns out that getencoding was added on 3.11 :( https://docs.python.org/3/library/locale.html?highlight=locale#locale.getencoding

Maybe we can just ignore all this ad stick utf8 there, but I fear we may have some windows user complain.
@zzzeek do you think we can just assume utf8?

kkirsche:
Worst case, my suggestion would be that we add a comment to the line above this indicating why we aren't / can't reasonably specify an encoding. That way we have a reference to where we got stuck with this in case you receive additional requests about this or while future refactoring work is done. 

Thank you both again for your time, help and consideration. I know you guys volunteer to do this work, so I appreciate you sharing your time with me for this.

CaselIT:
that may be enough. Let's wait for mike opinions on this before proceeding

zzzeek:
this warning mode was only added in python 3.11.

therefore in this case we make a new alembic/util/compat.py function `read_config_parser()`:

```py
def read_config_parser(file_config, file_argument):
    if py311:
                                                                             # can this be `encoding="locale"` ?  the pep is not clear
        return file_config.read(file_argument, encoding=locale.getencoding())
  
   else:
        return file_config.read(file_argument)
```

what are the risks of `locale.getencoding()` having a problem ?  returning something nonsensical that is not what actually gets used if we leave encoding=None ?

kkirsche:
Great questions, and I'll do my best to answer from my understanding and assumptions.

First:

> this warning mode was only added in python 3.11.

This is not correct. It was added in 3.10 per https://docs.python.org/3/library/io.html#opt-in-encodingwarning. That changes the compatibility range slightly. This also coincides with when the `"locale"` value became supported as an `encoding` argument per https://docs.python.org/3/whatsnew/3.10.html#optional-encodingwarning-and-encoding-locale-option

> what are the risks of `locale.getencoding()` having a problem ? returning something nonsensical that is not what actually gets used if we leave encoding=None ?

Based on the implementation written in https://github.com/sqlalchemy/alembic/pull/1274#discussion_r1263835114 there is no additional risk as they are identical behaviors, though one is explicit and removes the warning, and one is implicit.

The `io.text_encoding` can only return `"locale"` or `"utf-8"` when encoding is not set. Given that `"locale"` and the warning both were released in Python 3.10, my suggestion for a compatibility function would be:


```python
def read_config_parser(file_config, file_argument):
    if py310:
        return file_config.read(file_argument, encoding="locale")
  
    else:
        return file_config.read(file_argument)
```

This bypasses the concerns about `locale.getencoding` and instead relies on the behaviors released in Python 3.10+

kkirsche:
Corrected my statement in an edit if you only see this via email. Sorry about that

zzzeek:
> Great questions, and I'll do my best to answer from my understanding and assumptions.
> 
> First:
> 
> > this warning mode was only added in python 3.11.
> 
> This is not correct. It was added in 3.10 per https://docs.python.org/3/library/io.html#opt-in-encodingwarning. That changes the compatibility range slightly. This also coincides with when the `"locale"` value became supported as an `encoding` argument per https://docs.python.org/3/whatsnew/3.10.html#optional-encodingwarning-and-encoding-locale-option

oh.   that part is 3.10.  but `getencoding()` is 3.11.  so that's bad.

> Given that `"locale"` and the warning both were released in Python 3.10, my suggestion for a compatibility function would be:
> 
> ```python
> def read_config_parser(file_config, file_argument):
>     if py310:
>         return file_config.read(file_argument, encoding="locale")
>   
>     else:
>         return file_config.read(file_argument)
> ```
> 
> This bypasses the concerns about `locale.getencoding` and instead relies on the behaviors released in Python 3.10+

where is `encoding="locale"` documented?  

nope: https://docs.python.org/3/library/configparser.html?highlight=configparser#configparser.ConfigParser.read
nope: https://docs.python.org/3/library/functions.html#open
nope: https://docs.python.org/3/library/codecs.html#module-codecs

how did you find this?

zzzeek:
> To use the current locale encoding, encoding="locale" is supported since Python 3.10.

OK so version check for py310 then encoding="locale", thanks

kkirsche:
> how did you find this?
> 
> 
> 
> 
> 
> **update** seems to be here: https://docs.python.org/3/library/io.html#text-encoding

For new features I don't bother going through the libraries looking. I've found it much faster to find the version release notes and then go from their feature lists. This often directs you to the correct location which is where you saw above. 

So here I saw:

https://docs.python.org/3/whatsnew/3.10.html#optional-encodingwarning-and-encoding-locale-option

Which links to:
https://docs.python.org/3/library/io.html#io-text-encoding

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 5e5e1c1156c1fd00e447defa758459bc41b24b9f of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 5e5e1c1156c1fd00e447defa758459bc41b24b9f added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4786

zzzeek:
looks like some new commits were added.  send a message here when next to re-run CI

kkirsche:
> looks like some new commits were added.  send a message here when next to re-run CI

I'll need to see if I can figure out what Jenkins is upset about. From the output it didn't seem to be my commits, but maybe I'm just not used to navigating around Gerrit and Jenkins

zzzeek:
looks like it's failing for pep8

kkirsche:
I believe that's been fixed in fac1b42. `pycodestyle`

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision fac1b42bf3df457d7105d80513b33aee35a611bb of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset fac1b42bf3df457d7105d80513b33aee35a611bb added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4786

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision d0abd33845754827b8ce69ba1360989c7c57fdb6 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset d0abd33845754827b8ce69ba1360989c7c57fdb6 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4786

sqla-tester:

fixed the linter fails

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4786

kkirsche:
Sorry, I couldn't find in the docs how to run the linters as they didn't seem to run with the default tox suite, so I was relying on CI/CD for feedback. Sorry again for the inconvenience

CaselIT:
no problem!

kkirsche:
Thank you, zzzeek. If you need anything from me just let me know. I appreciate your time, patience and help.

PR #1368, opened 2023-11-29
Improve `Rewriter` implementation

### LINKED ISSUE #1337: Improve re-writer implementation, fixing chain and adding support to callables
ive just read the source code for chain() and it's a bit of a disaster.  because if i do this:

    w = writer1.chain(writer2).chain(writer3)

writer2 is lost.  it's doing the chaining wrong.  my high school comp sci teacher would be unimpressed

it works for your use case right now because it only handles exactly two writers.   but if we fix it to correctly implement chaining then it would not work for your case in the general sense.    

we can fix both issues at once by 1. fixing the chaining so that if `self._chain` is non-None we traverse the list through to the end for the linked list append, and 2. looking at the type of object sent and if it's just a callable, wrapping it in a helper object `CallableRewriter` that uses that callable and also supports the rest of the `Rewriter` protocol including `self._chain`.

_Originally posted by @zzzeek in https://github.com/sqlalchemy/alembic/discussions/1332#discussioncomment-7369839_

zrotceh:
Fixes #1337

### Description
* Fix the chaining of more than two rewriters
* Wrap a callable so that it could be chained

This works in my local test, and I hope it makes sense to use the callable wrapper as the base class.

### Checklist
This pull request is:

- [x] A short code fix
	- please include tests.


sqla-tester:
New Gerrit review created for change d43b320371d69798e2ad7bbf66c01daa2064ba55: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990

sqla-tester:
**Michael Bayer** (zzzeek) wrote:

this is probably OK, need to look more closely can we get a changelog file that indicates to users what this change is?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990

alembic/autogenerate/rewriter.py
```diff
@@ -23,41 +23,40 @@
     from ..runtime.environment import _GetRevArg
     from ..runtime.migration import MigrationContext
 
+    ProcessRevisionDirectiveFn = Callable[
```
sqla-tester, line 26: **Federico Caselli** (CaselIT) wrote:

It probably makes sense to move the type aliases outside of the type checking if

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990

alembic/autogenerate/rewriter.py
```diff
@@ -23,41 +23,40 @@
...
-    _chained: Optional[Rewriter] = None
+    _chained: Optional[RewriterBase] = None
 
-    def __init__(self) -> None:
-        self.dispatch = util.Dispatcher()
+    def __init__(
+        self,
+        process_revision_directives: Optional[
+            ProcessRevisionDirectiveFn
+        ] = None,
```
sqla-tester, line 48: **Federico Caselli** (CaselIT) wrote:

this could be required, meaning remove the `Optional` and the `=None`

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990

alembic/autogenerate/rewriter.py
```diff
@@ -89,9 +88,56 @@ def add_column_idx(context, revision, op):
         """
         wr = self.__class__.__new__(self.__class__)
         wr.__dict__.update(self.__dict__)
-        wr._chained = other
+        if not isinstance(other, RewriterBase):
+            other = RewriterBase(process_revision_directives=other)
+        final_writer = wr
+        while final_writer._chained:
+            final_writer = final_writer._chained
+        final_writer._chained = other
```
sqla-tester, line 96: **Federico Caselli** (CaselIT) wrote:

I'm not sure this is the best options, since this changes a chained rewriter if it's parted is chained again. I mean that


    # r0, r1, r2 are rewriters
    r3 = r1.chain(r2) # r1 and r2 unchanged. r3._chainedis r2
    r4 = r3.chain(r0) # r2 is changed 
  
Maybe it's easier to change `_chained` to be a tuple that starts as an empty tuple (haven't tried it but I think it should work)?
At that point `RewriterBase` is no longer needed, and we can do here
```py
wr._chained += (other,)
```

it's fine for other to be a callable or another rewriter, The call method could then loop over _chained and call all items in the tuple

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990

sqla-tester (commented):

Thanks for the PR!

List a suggesting that I think would make for an easier implementation and would avoid some unwanted modification of chained rewriters

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990

alembic/autogenerate/rewriter.py
zrotceh: I got a `NameError: name 'MigrationContext' is not defined` when doing so.

alembic/autogenerate/rewriter.py
zrotceh: I believe this no longer applies.

alembic/autogenerate/rewriter.py
zrotceh: PR updated, thanks for the suggestion.

alembic/autogenerate/rewriter.py
CaselIT: You can make that a string, similar to how migration script is

zrotceh: Ah, I see. Thank you!

CaselIT:
Thanks for the update. Let me update gerrit!

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision e62633dbbb9b8dd91a145a6f27efcdbe4e4c0b3b of this pull request into gerrit so we can run tests and reviews and stuff

alembic/autogenerate/rewriter.py
```diff
@@ -23,6 +23,10 @@
     from ..runtime.environment import _GetRevArg
     from ..runtime.migration import MigrationContext
 
+ProcessRevisionDirectiveFn = Callable[
```
sqla-tester, line 26: **Federico Caselli** (CaselIT) wrote:

Done

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990

sqla-tester (commented):

Looks fine to me. Mike can you review?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990

* alembic/autogenerate/rewriter.py (line 48): Done
* alembic/autogenerate/rewriter.py (line 96): Done

sqla-tester:
**Michael Bayer** (zzzeek) wrote:

great!  thanks

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/4990 has been **merged**. Congratulations! :)

Issue #1386, opened 2024-01-05
Autogenerate renders TypeDecorator instance instead of underlying impl type


saifelse:
**Describe the bug**

This isn't a bug _per se_, but a small improvement for autogenerate when using TypeDecorator.

When a TypeDecorator is used in a column definition, e.g.:

```py
"""
File: app/models/foo.py
"""
from sqlalchemy.dialects.postgresql import JSONB
from sqlalchemy.types import TypeDecorator
...

class JSONBData(TypeDecorator):
   impl = JSONB

foo = Table("foo", MetaData(), Column("data", JSONBData))
```

The auto-generated migration ends up referencing the TypeDecorator:

```py
op.add_column("foo", sa.Column("data", app.models.foo.JSONBData(), nullable=True))
```

which is annoying for two reasons:

1. The import is not automatically rendered.
2. The migration file has an unnecessary dependency on `app`, e.g. if the app/models/foo.py is refactored, we may need to update this migration file... when that could have been avoided if instead of rendering `app.models.foo.JSONBData`, alembic directly rendered the underlying impl: `postgresql.JSONB`.

I'm not sure if there are any scenarios where it is actually preferable to have the TypeDecorator in the autogenerated file. If there are use cases for it, would it be sensible to make this a config option instead of unconditional?

**Expected behavior**
Ideally, we'd generate the same thing as when `foo = Table("foo", MetaData(), Column("data", JSONB))` is provided, i.e.:

```py
op.add_column("foo", sa.Column("data", postgresql.JSONB(astext_type=Text()), nullable=True))
```

**To Reproduce**

Test case:

```diff
diff --git a/tests/test_autogen_render.py b/tests/test_autogen_render.py
index eeeb92e..9755869 100644
--- a/tests/test_autogen_render.py
+++ b/tests/test_autogen_render.py
@@ -33,6 +33,7 @@ from sqlalchemy.sql import false
 from sqlalchemy.sql import literal_column
 from sqlalchemy.sql import table
 from sqlalchemy.types import TIMESTAMP
+from sqlalchemy.types import TypeDecorator
 from sqlalchemy.types import UserDefinedType
 
 from alembic import autogenerate
@@ -1078,6 +1079,21 @@ class AutogenRenderTest(TestBase):
             "server_default='5', nullable=True))",
         )
 
+    def test_render_add_column_type_decorator(self):
+        self.autogen_context.opts["user_module_prefix"] = None
+
+        class MyType(TypeDecorator):
[...]

PR #1387, opened 2024-01-05
render type.impl when type is a TypeDecorator instance

saifelse:
### Description
When rendering the type, we check to see if we are dealing with a TypeDecorator, and if so, proceed but with type.impl instead... this results in the underlying type from the sqlalchemy package to be rendered into the migration instead of the TypeDecorator (wherever it may be defined).

Fixes: #1386

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/autogenerate/render.py
+++ b/alembic/autogenerate/render.py
@@ -834,7 +834,7 @@ def _repr_type(
             prefix = _sqlalchemy_autogenerate_prefix(autogen_context)
             return "%s%r" % (prefix, type_)
     elif isinstance(type_, sqltypes.TypeDecorator):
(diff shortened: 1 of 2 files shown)


zzzeek:
as mentioned in my [comment](https://github.com/sqlalchemy/alembic/issues/1386#issuecomment-1878667216) I dont think we want to do this.  We definitely get users complaining etc. when their exact type is not rendered, even if they don't need it to be.    This seems like a big leap of assumptions to make for not much reason, not the least of which that the column as returned inside the `Table` from a `create_table()` directive will not have their preferred type.

As mentioned in my comment, this particular preference can be suited by a simple type rendering recipe right now for those who really want to control this.

PR #1397, opened 2024-01-13
Fix `alembic.ini` templates to match `configparser` file format.

bhushan-mohanraj:
In the `alembic.ini` templates, I moved the inline comment about `version_path_separator` to their own lines as required by `configparser`.

### Description

In a recent project, I included the following configuration values in my `alembic.ini`. Note that the last line is the default line from the current generic `alembic.ini` template.

```
# version location specification; This defaults
# to migrations/versions.  When using multiple version
# directories, initial revisions must be specified with --version-path.
# The path separator used here should be the separator specified by "version_path_separator" below.
version_locations = %(here)s/migrations/versions

# version path separator; As mentioned above, this is the character used to split
# version_locations. The default within new alembic.ini files is "os", which uses os.pathsep.
# If this key is omitted entirely, it falls back to the legacy behavior of splitting on spaces and/or commas.
# Valid values for version_path_separator are:
#
# version_path_separator = :
# version_path_separator = ;
# version_path_separator = space
version_path_separator = os  # Use os.pathsep. Default configuration used for new projects.
```

When running `alembic check`, I encountered:

```
ValueError: 'os  # Use os.pathsep. Default configuration used for new projects.' is not a valid value for version_path_separator; expected 'space', 'os', ':', ';'
```

It seemed that the comment in the last line was being included as part of the parsed config value, which should be `os`.

Alembic currently [uses `configparser.ConfigParser` from the standard libary](https://github.com/sqlalchemy/alembic/blob/2d60c77c81a72a78b575b96aef511e658073dec5/alembic/util/compat.py#L82-L89) to parse `alembic.ini` files. The [default `configparser` file format](https://docs.python.org/3/library/configparser.html#supported-ini-file-structure) requires that comments be on their own lines, although this can be customized. I changed the three copies of this line in Alembic's `alembic.ini` templates to remove the inline comments.

This issue could also be fixed by changing [the default instance of `ConfigParser`](https://github.com/sqlalchemy/alembic/blob/2d60c77c81a72a78b575b96aef511e658073dec5/alembic/config.py#L202), using `inline_comment_prefixes=("#",)`. I imagine, however, that it might be better to use the default file format.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/templates/async/alembic.ini.mako
+++ b/alembic/templates/async/alembic.ini.mako
@@ -47,7 +47,9 @@ prepend_sys_path = .
 # version_path_separator = :
 # version_path_separator = ;
 # version_path_separator = space
-version_path_separator = os  # Use os.pathsep. Default configuration used for new projects.
+#
+# Use os.pathsep. Default configuration used for new projects.
+version_path_separator = os
 
 # set to 'true' to search source files recursively
 # in each "version_locations" directory
--- a/alembic/templates/generic/alembic.ini.mako
+++ b/alembic/templates/generic/alembic.ini.mako
@@ -49,7 +49,9 @@ prepend_sys_path = .
 # version_path_separator = :
 # version_path_separator = ;
 # version_path_separator = space
-version_path_separator = os  # Use os.pathsep. Default configuration used for new projects.
+#
+# Use os.pathsep. Default configuration used for new projects.
+version_path_separator = os
 
 # set to 'true' to search source files recursively
 # in each "version_locations" directory
--- a/alembic/templates/multidb/alembic.ini.mako
+++ b/alembic/templates/multidb/alembic.ini.mako
@@ -49,7 +49,9 @@ prepend_sys_path = .
 # version_path_separator = :
 # version_path_separator = ;
 # version_path_separator = space
-version_path_separator = os  # Use os.pathsep. Default configuration used for new projects.
+#
+# Use os.pathsep. Default configuration used for new projects.
+version_path_separator = os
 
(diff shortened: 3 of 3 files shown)


CaselIT:
Thanks

Seems to be a reasonable change

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision f7198e39aa077218a01a213c5204a0dd6502b3aa of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Failed to create a gerrit review, git squash against branch 'main' failed

zzzeek:
I need this PR to be rebased on main for me to work with it.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 073cbd9b076fffa78be3b374e20ece4a2562f7bd of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 073cbd9b076fffa78be3b374e20ece4a2562f7bd: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5587

zzzeek:
OK I used github's UX for this

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5587 has been **merged**. Congratulations! :)

PR #1529, opened 2024-09-03
Mention time zone data unavailability on some systems

floxay:
As discussed in https://github.com/sqlalchemy/alembic/discussions/1528

### Description
Mentions that on some systems time zone data might be unavailable and an additional dependency on `tzdata` will fix the errors in this case.

### Checklist
This pull request is:

- [ ] A short code fix
	- please include tests.


CaselIT:
> Any required deps can installed by adding `alembic[tz]` to the pip requirements

I wonder if the tz extra should include tzdata. It's required only in some cases, so not sure.
It also seems that it depends on the python distribution (anaconda vs official one) so it seems like a bit of a mess.

@zzzeek what do you think?

zzzeek:
this is superseded by #1558

PR #1541, opened 2024-09-22
Add colon (:) to invalid char for revision name

### LINKED ISSUE #1540: Colon (:) shouldn't be allowed in custom revision id, as it is used as revision range
**I created a revision named REV:1, when I tried to upgrade to it using**
`alembic upgrade REV:1`
**It showed error**
`FAILED" Range revision not allowed`

**Expected: Don't allow revision name containing colon**

PatilHrushikesh:
FIXES #1540 

### Description
As colon is used in revision range, it shouldn't be used in revision name.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/script/revision.py
+++ b/alembic/script/revision.py
@@ -45,7 +45,7 @@ _T = TypeVar("_T")
 _TR = TypeVar("_TR", bound=Optional[_RevisionOrStr])
 
 _relative_destination = re.compile(r"(?:(.+?)@)?(\w+)?((?:\+|-)\d+)")
-_revision_illegal_chars = ["@", "-", "+"]
+_revision_illegal_chars = ["@", "-", "+", ":"]
 
 
 class _CollectRevisionsProtocol(Protocol):


zzzeek:
it's too bad because the colon isn't any problem for the revision API, it's just used by the command interface.  colon could be supported if the command interface had a function that parsed for quoted revision names.  but then we'd probably want quotes added to the illegal chars also

PatilHrushikesh:
### Can you explain
> it's just used by the command interface.

Yeah, and it's causing issue mentioned in ticket, is there better way to fix it?



>  quoted revision

How does quoted revision comes into the picture?

CaselIT:
I think the current solution is likely easier to implement. What do you think mike?

zzzeek:
it is.  can you add a changelog and do we have illegal char tests set up?  we should

CaselIT:
@PatilHrushikesh 
- could you add an additional case (or edit an existing one) here https://github.com/sqlalchemy/alembic/blob/9cc68f5e89183e3c5634ac892bf7819dfc90a8d5/tests/test_script_production.py#L347-L390 to have a `:`?
- could you add a changelog entry here? examples are here https://github.com/sqlalchemy/alembic/tree/3db374f19760320507a96443e29d19835654b2ec/docs/build/unreleased the number is the issue id

Thanks!

zzzeek:
is this PR stalled ?   it needs the additional change as well as a unit test

zzzeek (changes requested):
any chance for unit tests here?

zzzeek:
superseded by #1741 ; i gave both authors credit

Issue #1560, opened 2024-10-30
Make version table customizable by dialects


maver1ck:
**Describe the use case**
Starrocks is Open-Source, High-Performance Analytical Database.
https://www.starrocks.io

**Databases / Backends / Drivers targeted**

Starrocks

**Example Use**
Management of Tables in Starrocks.

**Additional context**
I will create PR with detailed technical solution.

**Have a nice day!**

zzzeek:
Alembic only includes support for SQLAlchemy's built-in dialects: oracle, postgresql, sqlite, mysql, mssql.    To add alembic implementations to third party dialects, you should contribute your PR to that dialect's github repository.

zzzeek:
the alembic_version table has only one column.  it is by definition the first column in the table.

zzzeek:
both statements are identical

```py
>>> from sqlalchemy import table, column
>>> alembic_version = table("alembic_version", column("version_num"))
>>> print(alembic_version.select())
SELECT alembic_version.version_num 
FROM alembic_version
>>> print(alembic_version.select().with_only_columns(alembic_version.c.version_num))
SELECT alembic_version.version_num 
FROM alembic_version

```

zzzeek:
this is why, let's open a discussion and step through what's happening

CaselIT:
I think the issue is that they want to use a table that has two columns with an autoincrement id as the first column. Not sure if it's a constraint of starrocks the fact that it requires one such column as the fist one.

sqla-tester:
**Maciek Bryński** has proposed a fix for this issue in the **main** branch:

**Add ability to configure alembic_version table in DialectImpl** https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5556

zzzeek:
ok you're released thanks for the patience

maver1ck:
Thanks @zzzeek.

jaogoy:
@maver1ck  Hi friend, have it solved the problem?
Currently, I've tested alembic with StarRocks, I still encounter the problem with alembic_version. (of course, there are also a lot of other problems.)

maver1ck:
Hi @jaogoy,
I'm using this implementation of Dialect along with Starrocks with success.
https://github.com/StarRocks/starrocks/blob/main/contrib/starrocks-python-client/starrocks/alembic.py#L10

jaogoy:
[...]

PR #1561, opened 2024-10-30
Initial work to support Starrocks DB

maver1ck:
I want to use Alembic with Starrocks database.

To achieve this I need to patch `alembic_version` table with following code

```
def patch_alembic_version(context, **kwargs):
    migration_context = context._proxy._migration_context
    old_version = migration_context._version
    metadata = old_version.metadata
    metadata.remove(old_version)

    version = Table(
        'alembic_version',
        metadata,
        Column('id', types.BIGINT, autoincrement=True, primary_key=True),
        Column('version_num', types.VARCHAR(32), primary_key=False),
        starrocks_primary_key="id"
    )
    migration_context._version = version
 ```

This solution works with one exception - select on alembic_version always read version_num from first column.

### Description
Fixes: https://github.com/sqlalchemy/alembic/issues/1560

### Checklist
This pull request is:

- [X] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/runtime/migration.py
+++ b/alembic/runtime/migration.py
@@ -540,7 +540,7 @@ class MigrationContext:
                 return ()
         assert self.connection is not None
         return tuple(
-            row[0] for row in self.connection.execute(self._version.select())
+            row[0] for row in self.connection.execute(self._version.select().with_only_columns(self._version.c.version_num))
         )
 
     def _ensure_version_table(self, purge: bool = False) -> None:


zzzeek:
hi -

im not sure what problem you are having but this PR doesn't really make any sense, as it just adds a command to the `select()` here that would not really change anything since alembic_version already has just that one column.

Please open a [discussion](https://github.com/sqlalchemy/alembic/discussions) with a description of the actual problem you are having including complete error messages / stack trace and we can begin to see where your solution might be.

PR #1627, opened 2025-03-19
Add support of IF [NOT] EXISTS for ADD/DROP COLUMN in Postgresql

### LINKED ISSUE #1626: Add support of IF [NOT] EXISTS for ADD/DROP COLUMN in Postgresql
**Describe the use case**
Add support for `IF [NOT] EXISTS` for `ADDDROP COLUMN` operations available on PostgreSQL

PostgreSQL

**Example Use**
The goal is to have automatically generated operations rewritable to add `if_exists/if_not_exists` attributes, then rendered with postgres compiler.

```python

@writer.rewrites(ops.AddColumnOp)
def add_column(autogen_context, _revision, op: ops.AddColumnOp):
    op.if_not_exists = True
```

would generate following SQL
```sql
ALTER TABLE t ADD COLUMN IF NOT EXISTS c INTEGER;
```

**Additional context**
Spin-off of #524 for `ADD/DROP COLUMN`

**Have a nice day!**

lachaib:
Fixes #1626

### Description

Enable support of IF NOT EXISTS for AddColumn/DropColumn so that the flow from auto-generation to execution on PostgreSQL will work.

Took the opportunity to mention the `if_exists/if_not_exists` in a cookbook entry also for CreateTable/CreateIndex

Disclaimer: I'm still not fully proficient with `alembic`'s code so don't hesitate to challenge the implementation.

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**


alembic/autogenerate/render.py
```diff
@@ -457,40 +457,56 @@ def _drop_constraint(
 
 @renderers.dispatch_for(ops.AddColumnOp)
 def _add_column(autogen_context: AutogenContext, op: ops.AddColumnOp) -> str:
-    schema, tname, column = op.schema, op.table_name, op.column
+    schema, tname, column, postgresql_if_not_exists = (
+        op.schema,
+        op.table_name,
+        op.column,
+        op.postgresql_if_not_exists,
```
zzzeek, line 464: we can't have postgresql-specific logic in render here. but also, the op should not have a `.postgresql_if_not_exists` field anyway, it should be in `.kw`.    

while all of the ops have a `.kw` field that stores additional dialect-specific keyword arguments, these are not included in renders right now.    we can try adding these to renders but this would need tests across many 

that said it's not clear if this argument should be called `postgresql_if_not_exists` anyway

zzzeek (changes requested):
hi -

while we can't go with any of the "postgresql_" hardcoding here outside of postgresql.py, there are plenty of other databases that support IF NOT EXISTS/IF EXISTS for alter table add/drop column, just googling shows there are a bunch like snowflake, etc.

So please just remove this from being specific to postgresql and implement in a similar way as if_exists/if_not_exists is done for create_index/drop_index/create_table/drop_table etc.

alembic/operations/ops.py
```diff
@@ -2034,16 +2034,20 @@ def __init__(
         column: Column[Any],
         *,
         schema: Optional[str] = None,
+        postgresql_if_not_exists: Optional[bool] = None,
```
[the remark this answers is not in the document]
CaselIT: We already have precedent here, like postgresql_using https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.alter_column.params.postgresql_using

Overall if only postgresql suppots it maybe it makes sense to start with it then expand if there is a request for it?

CaselIT:
Replied above regarding the prefix

alembic/operations/ops.py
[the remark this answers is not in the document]
zzzeek: the word "postgresql" does not appear in the ops.py file outside of a docstring

we cannot have dialect-level rules hardcoded in the core of alembic full stop

the use case for this user is they explicitly want *rendering* to render this keyword

so that will not work, because we do not render the ``kw`` attribute of ops.  They can make their own `@render` function to override that but that's pretty inconvenient

in this specific case, "if exists / if not exists" is extremely common

[mariadb](https://mariadb.com/kb/en/alter-table/)  - that alone is reason enough
[snowflake](https://docs.snowflake.com/en/sql-reference/sql/alter-table)
[jooq has a whole list including bigquery, cockroachdb, databricks, teradata](https://www.jooq.org/doc/latest/manual/sql-building/ddl-statements/alter-statement/alter-table-statement/alter-table-add-column-if-not-exists/) - and they actually have syntaxes for Oracle, SQL Server too using short procedures, so we could even do those too

so in short this flag should be added without any prefix, however postgresql and mysql/mariadb impls can handle the rendering functions, need tests in test_postgresql test_mysql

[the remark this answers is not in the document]
CaselIT: ok, I guess the main difference here is that with postgresql_using we never expect it to be automatically rendered.

Regarding the availability in other db I missed that mariadb already had it.

lachaib:
@zzzeek 
OK so we're back to `is_exists` & `if_not_exists` and I added support for mariadb dialect in mysql, though according to what I saw mysql doesn't support it, so I did register only for mariadb. Is that ok?

zzzeek:
> @zzzeek OK so we're back to `is_exists` & `if_not_exists` and I added support for mariadb dialect in mysql, though according to what I saw mysql doesn't support it, so I did register only for mariadb. Is that ok?

I think it's best to register for both since use of the explicit "mariadb://" style URL is not common

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 1cd9a3968f7ffa27929eafafc40b36f9087b653c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 1cd9a3968f7ffa27929eafafc40b36f9087b653c: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5879

lachaib:
> > @zzzeek OK so we're back to `is_exists` & `if_not_exists` and I added support for mariadb dialect in mysql, though according to what I saw mysql doesn't support it, so I did register only for mariadb. Is that ok?
> 
> I think it's best to register for both since use of the explicit "mariadb://" style URL is not common

There you go with an updated version. Let's hope it suits you now :)

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 7702b01d69c4487d4569fe067a1e04ed4dc9b88d of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 7702b01d69c4487d4569fe067a1e04ed4dc9b88d added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5879

lachaib:
CI is getting failures that seem unrelated to my work, any help would be appreciated to help me go further

zzzeek:
> CI is getting failures that seem unrelated to my work, any help would be appreciated to help me go further

that looks as a direct result of the patch here, the error is here: https://jenkins.sqlalchemy.org/job/alembic_gerrit/pyv=py313,sqla=sqlamain/1374/testReport/junit/tests.test_batch/BatchRoundTripPostgresqlTest_postgresql+psycopg2_16_8/test_drop_col_schematype/

```
File "/home/jenkins/workspace/alembic_gerrit/89e52482/sqlalchemy/alembic/alembic/ddl/impl.py", line 393, in drop_column
    base.DropColumn(
    ~~~~~~~~~~~~~~~^
        table_name,
        ^^^^^^^^^^^
    ...<2 lines>...
        **kw,
        ^^^^^
    )
    ^
TypeError: DropColumn.__init__() got an unexpected keyword argument 'existing_type'
```

that is code that was added by your patch on line 392:

https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5879/2/alembic/ddl/impl.py#392

zzzeek:
federico also made a comment on that code that maybe `**kw` should not be added there and I agree it seems like that would need to take bigger things into account

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 7154331533170b16ed89d27413eefc4dee11d5b9 of this pull request into gerrit so we can run tests and reviews and stuff

zzzeek:
once the PR is working maybe I can go and see why kw isnt compatible there

sqla-tester:
Patchset 5391908789068c0204b93f962fb75351f039310f added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5879

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 5391908789068c0204b93f962fb75351f039310f of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 5391908789068c0204b93f962fb75351f039310f added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5879

zzzeek:
hi -

we've requested changes here but I'm not sure they propagated to the PR.  can you take a look at the comments listed at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5879?tab=comments and make sure all those changes are made here?  thanks

lachaib:
@zzzeek thanks for pointing me to gerrit I'm not used to it so I missed them, now all points there should be fixed, let me know

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 82d6319410b55e8fb3aa3c63c988b00f5d7721ba of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 82d6319410b55e8fb3aa3c63c988b00f5d7721ba added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5879

alembic/ddl/mssql.py
```diff
@@ -222,7 +223,9 @@ def drop_column(
         drop_fks = kw.pop("mssql_drop_foreign_key", False)
         if drop_fks:
             self._exec(_ExecDropFKConstraint(table_name, column, schema))
-        super().drop_column(table_name, column, schema=schema, **kw)
+        super().drop_column(
+            table_name, column, schema=schema, if_exists=if_exists, **kw
```
zzzeek, line 227: does mssql support DROP COLUMN IF EXISTS ? there's no test and wondering why this is here

alembic/ddl/mysql.py
```diff
@@ -391,6 +393,29 @@ def _mysql_alter_default(
     )
 
 
+@compiles(AddColumn, "mysql", "mariadb")
+def _mysql_add_column(
+    element: AddColumn, compiler: MySQLDDLCompiler, **kw
+) -> str:
+
+    return "%s ADD COLUMN %s%s" % (
```
zzzeek, line 401: how come this is done so differently vs. the postgresql version?

alembic/ddl/postgresql.py
```diff
@@ -512,6 +516,34 @@ def __init__(
         self.using = using
 
 
+@compiles(AddColumn, "postgresql")
+def visit_add_column(element: AddColumn, compiler: PGDDLCompiler, **kw) -> str:
+    return "%s %s" % (
+        alter_table(compiler, element.table_name, element.schema),
+        add_column(
```
zzzeek, line 523: why not just pass through if_exists / if_not_exists in base.py? it's just an argument, the compiler can still ignore it if it chooses. i dont think we need new AddColumn / DropColumn @compiles here

alembic/ddl/postgresql.py
```diff
@@ -848,3 +880,39 @@ def _render_potential_column(
...
+    compiler: DDLCompiler,
+    column: Column[Any],
+    *,
+    if_not_exists: Optional[bool] = None,
+    **kw,
+) -> str:
+    text = "ADD COLUMN "
+    if if_not_exists:
+        text += "IF NOT EXISTS "
+
```
zzzeek, line 895: why not put this in [base.py](http://base.py/)? if I have a migration that is adding / dropping columns that may or may not exist already, can't it be assumed I'm using a database that supports this syntax? would I want if_exists / if_not_exists to be silently ignored if I passed it?

even if we dont want it to render we should likely keep this in [base.py](http://base.py/) anyway and then control for when we actually render IF [NOT] EXISTS

zzzeek (changes requested):
close. can we simplify the rendering implementation? it looks like it added different kinds of methods to postgresql, mysql, mssql, seems like it should be more consolidated

alembic/ddl/postgresql.py
lachaib: Given that alembic can be used to apply migration on several servers, my primary concern was that the automated generation of `if_exist` would possibly break for someone using a combination of postgresql and something else.

However you master way more than me the different usages of alembic and if you feel confident that we can move it all to `base.py` and remove it from implementations, I'd be happy to do so.
(It'll also solve most of the other questions)

alembic/ddl/mysql.py
lachaib: Main element of explanation: I privileged local style consistency, and the implem files are way different one from the other. Also I did postgresql way before and worked on this file at another moment, hence privileging local style over daring to harmonise the contribution and make this change seem odd compared to the rest of the file

alembic/ddl/mssql.py
lachaib: it doesn't, but I had a failure with mypy without it (signature of override differing from parent class)

several options come to my mind:
* leave it so; mypy and linters happy, but can be misleading
* leave if_exists in signature, but don't pass it through; mypy is happy other linters may complain
* leave if_exists in signature, log a warning if not none, don't pass it through; linters are happy, users have a proper feedback of incorrect use
* remove it from signature and make add `# type: ignore` comment; may silently impact any future change on this API
* update parent signature to make the `if_exist` argument implicitly taken from `**kw`; mypy and linters happy, users and maintainer are not

My preferred option is the 3rd and I'm gonna submit an additional commit in this direction, we'll amend eventually if you select another one

alembic/ddl/postgresql.py
zzzeek: but to be clear, just do the first part of that paragraph unless you have a great idea for the second part.  otherwise Ill take the succinct version of the code and figure something out.

alembic/ddl/postgresql.py
lachaib: I pushed the move to base.py as a distinct commit for you to validate, I'll squash changes prior to merging if you confirm it's ok for you.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision c503b049f453a12e7fdb87464606cddd58ad306c of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset c503b049f453a12e7fdb87464606cddd58ad306c added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/5879

PR #1645, opened 2025-04-12
Add basic support for Starrocks

phaethon:
### Description
Starrocks is an analytical database (https://www.starrocks.io). It has its own dialect defined in sqlalchemy. Currently, alembic does not support this dialect. Starrocks implements MySQL protocol and many BI tools are recommended to use MySQL driver/connector to use Starrocks. This commit PR just subclasses MySQL driver to enable basic Starrocks support.

### DIFF
--- a/alembic/ddl/mysql.py
+++ b/alembic/ddl/mysql.py
@@ -321,6 +321,8 @@ class MySQLImpl(DefaultImpl):
 class MariaDBImpl(MySQLImpl):
     __dialect__ = "mariadb"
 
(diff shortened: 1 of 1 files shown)


zzzeek:
hi -

this code should go into the Starrocks dialect, not Alembic.      See https://github.com/sqlalchemy-redshift/sqlalchemy-redshift/blob/a077110366a3c451f1c41b2fa56f6e3cd5c438dd/sqlalchemy_redshift/dialect.py#L49 for an example with almost the exact same kind of dialect.

PR #1681, opened 2025-06-11
pass `context.config.config_args` to `logging.config.fileConfig`

jvanasco:
Some projects, like Pyramid, support `%(here)s` interpolation in `.ini` files for logging.

Extending Alembic to support this is trivial.  The data is already computed as part of `config.config_args`.  The `env.py` files just need to pass in these args as `defaults`:

-     fileConfig(config.config_file_name)
+     fileConfig(config.config_file_name, defaults=config.config_args)

### DIFF
--- a/alembic/templates/async/env.py
+++ b/alembic/templates/async/env.py
@@ -14,7 +14,7 @@ config = context.config
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
 if config.config_file_name is not None:
-    fileConfig(config.config_file_name)
+    fileConfig(config.config_file_name, defaults=config.config_args)
 
 # add your model's MetaData object here
 # for 'autogenerate' support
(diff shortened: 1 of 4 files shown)


zzzeek:
how is this fitting into what you're doing ?    is this for paths to log files?    all the alembic script locations etc. I think are already consumed by the time you get to env.py

jvanasco:
> is this for paths to log files?

Yes.

CaselIT:
if we do change this part, maybe we could add also the other flag suggested here? https://github.com/sqlalchemy/alembic/discussions/1483#discussioncomment-9920270

jvanasco:
> if we do change this part, maybe we could add also the other flag suggested here? https://github.com/sqlalchemy/alembic/discussions/1483#discussioncomment-9920270

A quick online search suggests that may be a good idea.  It seems to cause issues on many projects.

An alternative for both would just be putting some inline comments on the generated env.py file, pointing people how to do these things.

CaselIT:
maybe we could add both providing inline docs to the rationale about the args?
Something like
```py
fileConfig(
  config.config_file_name,
  # used for ...
  defaults=config.config_args,
  # ...
  disable_existing_loggers=False,
)

PR #1762, opened 2025-12-10
[Fixes #1761] Add opt-in autogenerate support for check constraints

### LINKED ISSUE #1761: Add opt-in autogenerate support for check constraints
**Describe the use case**

Currently, Alembic's autogenerate does not detect check constraints. When a CheckConstraint is added to or removed from a SQLAlchemy model, running alembic revision --autogenerate does not produce any migration operations for these constraints. This forces developers to  manually write migrations for check constraints, which is error-prone and inconsistent with how other constraints (unique,  foreign key) are handled.

This feature request is for an opt-in configuration option `compare_check_constraints`:
* Check constraints present in models but missing from the database; generates `CreateCheckConstraintOp`
* Check constraints present in the database but missing from models; generates `DropConstraintOp`

The comparison is done by name only to avoid false positives from SQL text normalization differences between databases and SQLAlchemy - this should be reasonably safe, as changing the text should mean that the constraint will be different in purpose and thus deserves a different name.

Any database that supports Inspector.get_check_constraints().

**Example Use**

```python
# configuration in env.py
context.configure(
    connection=connection,
    target_metadata=target_metadata,
    compare_check_constraints=True,
)

# model code
CheckConstraint("amount >= 0", name="ck_order_amount_non_negative")

# generated migration
def upgrade():
    op.create_check_constraint(
        'ck_order_amount_non_negative',
        'order',
        'amount >= 0'
    )
```

**Additional context**

* opt-in: will not change existing functionality
* throws if constraint is unnamed: otherwise no reliable way to match
* name-based comparison: avoiding false positives for SQL normalization would be difficult, especially for each dialect.
* filtered out auto-generated psql enum constraints

**Have a nice day!**

jrmalin:
### Description

This adds an opt-in compare_check_constraints configuration option for autogenerate, allowing Alembic to detect check constraint differences between SQLAlchemy models and the database.

Changes:
* New config option compare_check_constraints
* Check constraint comparison logic
* Check constraint signature class
* Renderer for CreateCheckConstraintOp
* Comparison method
* Type stub
* Tests covering add, remove, match, disabled, unnamed, invalid names, enum filtering, and bound parameters

### Checklist

This pull request is:

- [ ] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/autogenerate/compare.py
+++ b/alembic/autogenerate/compare.py
@@ -364,6 +364,18 @@ def _make_foreign_key(
     return const
 
 
+def _make_check_constraint(
+    params: Dict[str, Any], conn_table: Table
+) -> sa_schema.CheckConstraint:
+    const = sa_schema.CheckConstraint(
+        text(params["sqltext"]),
+        name=params["name"],
+        _autoattach=False,
+    )
+    const._set_parent_with_dispatch(conn_table)
+    return const
+
+
 @contextlib.contextmanager
 def _compare_columns(
     schema: Optional[str],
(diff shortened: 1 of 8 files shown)


jrmalin:
This has been released as part of [alembic-utils-extended](https://github.com/candidhealth/alembic-utils-extended), a fork of alembic_utils!

zzzeek:
this PR was discussed at #1761

PR #1794, opened 2026-02-21
Add --splice support to merge command

### LINKED ISSUE #1712: Merge command hints to do a slice, but there is no option for this
if it were me looking at this, I'd say the only problem is that merge is outputting that particular error message which comes from the "create a revision" mechanics and it should be reworded in that codepath to say "this is not supported by merge".    without looking into potential use cases here.

but then you are saying you wrote it all out as rev ids manually and "alembic is good enough"....which implies you're actually looking to make a merge point that **adds** a new head, rather than resolving two existing heads, and that....is beyond what I'd understand is actually useful in real world scenarios.     github is having an outage so I cant search for if there was an issue to add ``--splice`` to merge.   but at the same time it looks like ``--splice`` just disables a certain check from happening.    But overall I dont know what happens if you tell it to make a new revision with multiple bases and those bases aren't heads, I guess it just does it?     Here's the patch, it wont hurt on this end since it's nothing so feel free to try

```diff
diff --git a/alembic/command.py b/alembic/command.py
index 8e48547..ff9b159 100644
--- a/alembic/command.py
+++ b/alembic/command.py
@@ -385,6 +385,7 @@ def merge(
     message: Optional[str] = None,
     branch_label: Optional[_RevIdType] = None,
     rev_id: Optional[str] = None,
+    splice: bool = False,
 ) -> Optional[Script]:
     """Merge two revisions together.  Creates a new migration file.
 
@@ -435,6 +436,7 @@ def merge(
         refresh=True,
         head=revisions,
         branch_labels=branch_label,
+        splice=splice,
         **template_args,  # type:ignore[arg-type]
     )

```

_Originally posted by @zzzeek in https://github.com/sqlalchemy/alembic/discussions/1707#discussioncomment-14083857_

bysiber:
Fixes #1712

Adds a `splice` parameter to the `merge()` command function, matching the existing support in `revision()`. This is wired through to `generate_revision()` and since `splice` is already registered in `_KWARGS_OPTS`, the `--splice` CLI flag is automatically available for `alembic merge`.

Without this, trying to merge non-head revisions raises an error that suggests using `--splice`, but the merge command didn't actually accept that flag.

The patch is exactly what @zzzeek proposed in the linked discussion. Test included that verifies both the error without `splice` and success with `splice=True`.

### DIFF
--- a/alembic/command.py
+++ b/alembic/command.py
@@ -385,6 +385,7 @@ def merge(
     message: Optional[str] = None,
     branch_label: Optional[_RevIdType] = None,
     rev_id: Optional[str] = None,
+    splice: bool = False,
 ) -> Optional[Script]:
     """Merge two revisions together.  Creates a new migration file.
 
@@ -398,6 +399,9 @@ def merge(
 
     :param rev_id: hardcoded revision identifier instead of generating a new
      one.
+
+    :param splice: if True, allow the merge to create a new branch point
+     even if the given revisions are not heads.
 
     .. seealso::
 
@@ -435,6 +439,7 @@ def merge(
         refresh=True,
         head=revisions,
         branch_labels=branch_label,
+        splice=splice,
         **template_args,  # type:ignore[arg-type]
     )
 
--- a/tests/test_command.py
+++ b/tests/test_command.py
@@ -276,6 +276,36 @@ finally:
 
         assert os.path.exists(rev.path)
 
+    def test_merge_cmd_splice(self):
+        """merge with --splice allows non-head revisions to be merged."""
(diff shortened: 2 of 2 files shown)


sqla-tester:
New Gerrit review created for change a0f9b6d929b7124fbc9c4c0d4d8191f57a42222d: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6606

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6606 has been **merged**. Congratulations! :)

PR #1795, opened 2026-02-22
Fix : TypeDecorator rendering to preserve SchemaType name/schema 

khanjan2708:
Fixes Issue : sqlalchemy/sqlalchemy#13140 

### Description
This pull request fixes Issue sqlalchemy/sqlalchemy#13140, where `sqlalchemy.Enum` (and other `SchemaTypes` like `Boolean`) lost their `name `and schema properties when augmented with a `sqlalchemy.types.TypeDecorator` during autogenerated migrations.

The Fix: I have introduced a `_type_repr` helper in `alembic/autogenerate/render.py` that specifically detects `TypeDecorator` instances. If the implementation of the decorator is a `SchemaType`, Alembic now manually ensures that the `name` and `schema` attributes are preserved in the rendered migration string.

Additionally, `_user_autogenerate_prefix` was updated to use `.get()` instead of direct dictionary access to prevent `KeyError` regressions in certain testing environments where the full autogenerate context might not be initialized.

### Checklist

This pull request is:

- [x] A short code fix
	- please include tests.

**Have a nice day!**

### DIFF
--- a/alembic/autogenerate/render.py
+++ b/alembic/autogenerate/render.py
@@ -698,7 +698,7 @@ def _uq_constraint(
 
 
 def _user_autogenerate_prefix(autogen_context, target):
-    prefix = autogen_context.opts["user_module_prefix"]
+    prefix = autogen_context.opts.get("user_module_prefix", None)
     if prefix is None:
         return "%s." % target.__module__
     else:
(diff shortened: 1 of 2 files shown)


zzzeek (changes requested):
hey there

thanks for this.  however I listed out the approaches in https://github.com/sqlalchemy/alembic/issues/1551#issuecomment-2420595115 as well as another one in https://github.com/sqlalchemy/alembic/issues/1551#issuecomment-2420603412 

this approach seems to use another approach, which is to get the repr string and then tries to guess how to add "name" into it, which I'd rather not do this this way.   The core issue is TypeDecorator.__repr__ is broken and item 3 in my comment at https://github.com/sqlalchemy/alembic/issues/1551#issuecomment-2420595115 is the best way to approach this.

that is, we should be changing `TypeDecorator.__repr__` first and foremost.  if we want to put an interim fix on the alembic side, then it should not use the `TypeDecorator.__repr__` output at all and should re-implement from scratch.  

most effective would be my item 3 change, which would place this on the SQLAlchemy side.  ill propose that now

zzzeek:
I'm going to fix this on the SQLAlchemy side and we no longer need this fix on the alembic side.  any chance you can address the test failures on your other PR at #1793 ?   that PR is stalled as it does not pass tests

khanjan2708:
> I'm going to fix this on the SQLAlchemy side and we no longer need this fix on the alembic side. any chance you can address the test failures on your other PR at #1793 ? that PR is stalled as it does not pass tests

Working on it too

zzzeek:
hi -

the sqlalchemy fix is merged, closing this

PR #1823, opened 2026-06-27
Don't emit DROP of version table in offline mode at base (#1822)

### LINKED ISSUE #1822: Stamp command with "base" emits a DROP if --sql mode is present and no head revisions are present
**Describe the bug**

```shell
$ alembic stamp --sql base
BEGIN;

DROP TABLE alembic_version;

COMMIT;
```

**Expected behavior**

**To Reproduce**

**Versions.**
 - Python: `3.14.5`
 - Alembic: `1.18.4`
 - SQLAlchemy: `2.0.51`
 - Database: Postgresql `18.4`
 - DBAPI: Psycopg `3.3.4`

imurodl:
Fixes #1822.

As diagnosed by @zzzeek in the issue, offline (`--sql`) mode emitted a `DROP TABLE alembic_version` when running `stamp` or `downgrade` to `base`, while online mode never drops the version table — an inconsistency dating to the version table's introduction.

This removes the offline-only `DROP` so both modes behave the same. The version table is still created when it does not exist; only the spurious offline drop is gone.

**Tests**
- Updated `UpgradeDowngradeStampTest::test_version_to_none` to assert the version table is no longer dropped in offline downgrade-to-base.
- Added `UpgradeDowngradeStampTest::test_sql_stamp_to_base_no_drop` regressing the exact scenario from the issue (`stamp --sql base`).
- Added a changelog entry under `docs/build/unreleased/`.

Full local run: 1763 passed (the only failures are pre-existing `test_post_write.py` hook tests unrelated to this change). `black` + `flake8` clean.

### DIFF
--- a/alembic/runtime/migration.py
+++ b/alembic/runtime/migration.py
@@ -639,9 +639,11 @@ class MigrationContext:
                         run_args=kw,
                     )
 
-        if self.as_sql and not head_maintainer.heads:
-            assert self.connection is not None
-            self._version.drop(self.connection)
+        # NOTE: offline ("--sql") mode intentionally does not emit a DROP
+        # of the version table when ending at base.  Online mode never drops
+        # the version table (e.g. ``downgrade base`` only deletes its row),
+        # so dropping it in offline mode only was an inconsistency present
+        # since the version table was first introduced.  See #1822.
 
     def _in_connection_transaction(self) -> bool:
         try:
--- a/docs/build/unreleased/1822.rst
+++ b/docs/build/unreleased/1822.rst
@@ -1,0 +1,10 @@
+.. change::
+    :tags: bug, commands
+    :tickets: 1822
+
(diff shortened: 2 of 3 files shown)


sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision 8403c7083d9a1acbb9410ec0fd92331eace5ca8b of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 8403c7083d9a1acbb9410ec0fd92331eace5ca8b: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6846

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6846 has been **merged**. Congratulations! :)

PR #1824, opened 2026-06-28
Avoid dropping version table for offline SQL base

### LINKED ISSUE #1822: Stamp command with "base" emits a DROP if --sql mode is present and no head revisions are present

w-Jessamine:
### Description

When offline SQL generation ends at `base`, Alembic emitted `DROP TABLE alembic_version`. This removes that offline-only table drop so SQL output matches online behavior: Alembic updates version rows but does not drop the version table.

Fixes #1822.

### Checklist

This pull request is:

- [x] A short code fix

### Validation

- `/tmp/alembic-pr-venv/bin/python -m pytest tests/test_command.py::UpgradeDowngradeStampTest -q`
- `/tmp/alembic-pr-venv/bin/python -m pytest tests/test_command.py -k stamp -q`
- `/tmp/alembic-pr-venv/bin/python -m pytest tests/test_command.py -q`
- `/tmp/alembic-pr-venv/bin/python -m compileall -q alembic/runtime/migration.py tests/test_command.py`

### DIFF
--- a/alembic/runtime/migration.py
+++ b/alembic/runtime/migration.py
@@ -639,10 +639,6 @@ class MigrationContext:
                         run_args=kw,
                     )
 
-        if self.as_sql and not head_maintainer.heads:
-            assert self.connection is not None
-            self._version.drop(self.connection)
-
     def _in_connection_transaction(self) -> bool:
         try:
             meth = self.connection.in_transaction  # type:ignore[union-attr]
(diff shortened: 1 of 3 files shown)


zzzeek:
dupe of #1823

PR #1826, opened 2026-07-05
Strip PostgreSQL type cast specifiers when comparing computed defaults

### LINKED ISSUE #1462: Consider Removing psql type specifiers when Comparing Computed Constructs
**Describe the bug**
I am using a Computed column in a Postgresql database and created a migration. The next migration I create with automigration returns a warning: `alembic\autogenerate\compare.py:1034: UserWarning: Computed default on table.search_vector cannot be modified`

The string generated for comparison by `_normalize_computed_default`:
`setweightto_tsvectorenglish,title,a`
`setweightto_tsvectorenglish::regconfig,title::text,a::char`

The code responsible for this:

https://github.com/sqlalchemy/alembic/blob/44965f05e91ee5d424d9dde6566650c1bf26b516/alembic/autogenerate/compare.py#L979-L986

**Proposed solution**
Remove groups of text starting with `::` by adding another regex substitution.

```py
 def _normalize_computed_default(sqltext: str) -> str: 
     """we want to warn if a computed sql expression has changed.  however 
     we don't want false positives and the warning is not that critical. 
     so filter out most forms of variability from the SQL text. 
     """ 
  
    normalized_sqltext = re.sub(r"[ \(\)'\"`\[\]]", "", sqltext).lower()
    return re.sub(r"::\w+", "", normalized_sqltext)
```

cycsmail:
Fixes: #1462. On PostgreSQL, reflected computed-column expressions come back with type cast specifiers (e.g. `::regconfig`, `::text`, `::char`) that aren't present on the metadata side, causing autogenerate to raise a false-positive "Computed default ... cannot be modified" warning even when the expressions are equivalent.

Extends `_normalize_computed_default()` to strip those `::<type>` specifiers after the existing whitespace/quote normalization, so the two sides compare equal. Added a unit test normalizing the cast-laden expression from the issue against the cast-free one (fails without the fix, confirmed via `git stash`), plus a changelog entry under `docs/build/unreleased/`.

`tests/test_postgresql.py` passes (70 passed) and `ruff check` is clean on the touched files.

### Checklist

This pull request is:

- [x] A short code fix

**Have a nice day!**

### DIFF
--- a/alembic/autogenerate/compare/server_defaults.py
+++ b/alembic/autogenerate/compare/server_defaults.py
@@ -21,6 +21,7 @@ from ...util import sqla_compat
 if TYPE_CHECKING:
     from sqlalchemy.sql.elements import quoted_name
     from sqlalchemy.sql.schema import Column
+    from sqlalchemy.engine import Dialect
 
     from ...autogenerate.api import AutogenContext
     from ...operations.ops import AlterColumnOp
@@ -48,7 +49,7 @@ def _render_server_default_for_compare(
         return None
 
 
-def _normalize_computed_default(sqltext: str) -> str:
+def _normalize_computed_default(sqltext: str, dialect: Dialect) -> str:
     """we want to warn if a computed sql expression has changed.  however
     we don't want false positives and the warning is not that critical.
     so filter out most forms of variability from the SQL text.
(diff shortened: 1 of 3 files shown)


CaselIT (changes requested):
The proposed change cannot be there since it postgresql specific.

Also in the postgresql impl there is already something similar implemented in `_cleanup_index_expr` that could be re-used for this

alembic/autogenerate/compare/server_defaults.py
```diff
@@ -55,7 +55,10 @@ def _normalize_computed_default(sqltext: str) -> str:
 
     """
 
-    return re.sub(r"[ \(\)'\"`\[\]\t\r\n]", "", sqltext).lower()
+    normalized_sqltext = re.sub(r"[ \(\)'\"`\[\]\t\r\n]", "", sqltext).lower()
+    # strip postgresql type cast specifiers, e.g. ``::regconfig``, ``::text``,
+    # which can appear inconsistently and cause false-positive warnings.
+    return re.sub(r"::\w+", "", normalized_sqltext)
```
CaselIT: this can't be here since this is generic for all databases.

cycsmail:
good catch. gated the cast stripping on the postgresql dialect so it no longer affects other backends, the normalize helper now takes the dialect and only strips ::casts when it's postgres.

cycsmail:
just checking in on this one, happy to rebase or adjust anything if that makes it easier to review

PR #1830, opened 2026-07-15
Fix bare except clause in alembic/util/messaging.py

koteshyelamati:
Replace bare `except:` with `except Exception:` in the `status()` context manager.

The `status()` context manager prints a status message and a completion/failure indicator. Its bare `except:` catches `BaseException`, meaning `KeyboardInterrupt` or `SystemExit` during the operation would trigger the failure message and then re-raise — but the intent is clearly to catch operational exceptions, not signals. `except Exception:` is the correct type here.

### DIFF
--- a/alembic/util/messaging.py
+++ b/alembic/util/messaging.py
@@ -58,7 +58,7 @@ def status(
     msg(status_msg + " ...", newline, flush=True, quiet=quiet)
     try:
         yield
-    except:
+    except Exception:
         if not quiet:
             write_outstream(sys.stdout, "  FAILED\n")
         raise


MohammedAlkindi (commented):
Ran this on Windows 11 (10.0.26200), CPython 3.13.13, against `main` at `c116cbc0`. Failing-test **set** is identical to base (13 tests, the `test_post_write` and `ScriptNamingTest` timezone clusters, both failing on `main` here) — no regressions, none fixed.

**One thing worth deciding deliberately before this lands, because it is a behaviour change rather than a pure lint fix.** The bare `except:` is inside a context manager that re-raises immediately:

```python
@contextmanager
def status(status_msg, newline=False, quiet=False):
    msg(status_msg + " ...", newline, flush=True, quiet=quiet)
    try:
        yield
    except:
        if not quiet:
            write_outstream(sys.stdout, "  FAILED\n")
        raise
```

Because it re-raises unconditionally, this is one of the few genuinely defensible bare excepts — it swallows nothing. What it catches is *everything*, deliberately, so that the status line gets closed out. Narrowing to `except Exception:` means `KeyboardInterrupt` and `SystemExit` no longer print `FAILED`, so a **Ctrl-C during a migration now leaves the status line dangling** at `Running upgrade ... ` with no terminator, where today it prints `FAILED`.

That may well be what you want — arguably Ctrl-C is not a failure and shouldn't be labelled one. But it is a user-visible change to interrupt handling, not a no-op cleanup, so it seemed worth surfacing rather than letting it ride in as an E722 fix.

If the intent is purely to satisfy the linter while keeping the behaviour, `except BaseException:` does that exactly. If the intent is that interrupts should stop being reported as failures, then this patch is right as written and might deserve a line in the changelog, since anyone scripting against that output would see it change.

Either way the one-line diff is correct in isolation and I have no objection to it landing — I just did not want the interrupt case to go unnoticed.

PR #1832, opened 2026-07-15
fix: handle Path objects in version_locations

### LINKED ISSUE #1366: Path is not represented in current version locations
### Discussed in https://github.com/sqlalchemy/alembic/discussions/1284

<sup>Originally posted by **harpener** July 27, 2023</sup>
Hi guys.

**Describe the bug**

Cannot generate migration in project with multiple heads because location paths do not match.

**Expected behavior**

Migration is generated.

**To Reproduce**

```
alembic revision --autogenerate --head "after@head" -m "my_migration"
alembic revision --autogenerate --head "before@head" -m "my_migration"
```

**Error**

```
FAILED: Path C:\...\alembic\after is not represented in current version locations
FAILED: Path C:\...\alembic\before is not represented in current version locations
```

**Versions.**
 - OS: Windows 11 Pro, 22H2, 22621.1992
 - Python: 3.8.17
 - Alembic: 1.11.1
 - SQLAlchemy: 1.4.48
 - Database: PostgreSQL 9.6.2

**Additional context**

I use two heads - `after` and `before`. I have following in `alembic.ini`:
```
version_locations = %(here)s/alembic/before %(here)s/alembic/after
```

I noticed that my Python returns following:
```
>>> import os.path

>>> os.path.abspath("")
'c:\\...'

>>> os.path.realpath("")
'C:\\...'
```

As you can see, the letter-case of drive letter is different which is where the problem comes from.

I haven't encountered such error in past, but I recently installed the OS anew and updated Python to its latest security fix, if it helps.

Do you have any idea what is wrong?

**Have a nice day!**
</div>

pctablet505:
Fixes #1366

Windows users with multiple version_locations have been hitting "Path ... is not represented in current version locations" when generating revisions, for example with `alembic revision --autogenerate --head "after@head" -m "my_migration"`. The original report traced this to drive-letter casing differences on Windows (`os.path.abspath("")` returning `c:\...` while `os.path.realpath("")` returns `C:\...`), and follow-up comments showed the same error appears when paths contain spaces.

The 1.16.0 release moved version-location matching to pathlib, which normalizes case and spaces correctly for string entries. But `ScriptDirectory` already accepts `os.PathLike` objects in `version_locations`, and those still pass through `coerce_resource_to_filename`, which only accepted `str` and interpreted colons as package resources. A `pathlib.Path` would therefore not normalize the same way as string entries and could fall out of the allowed-location check.

This change widens `coerce_resource_to_filename` to accept `Union[str, os.PathLike[str]]` and only runs the package-resource branch for plain strings. Path inputs now become `pathlib.Path(...)` directly and get the same `.absolute()` normalization as string entries, so drive-letter casing and spaces are handled consistently. String-based configs, including package-resource references like `my.package:versions`, keep their existing behavior.

I added a regression test to `NormPathTest` in `tests/test_script_production.py` that builds a `ScriptDirectory` from `Path` version_locations and verifies that `generate_revision` can match a `Path` version_path. I also ran `pytest tests/test_script_production.py::NormPathTest` and `mypy` on the changed module; both passed.

### DIFF
--- a/alembic/util/pyfiles.py
+++ b/alembic/util/pyfiles.py
@@ -51,7 +51,9 @@ def template_to_file(
             f.write(output)
 
 
-def coerce_resource_to_filename(fname_or_resource: str) -> pathlib.Path:
+def coerce_resource_to_filename(
+    fname_or_resource: Union[str, os.PathLike[str]]
+) -> pathlib.Path:
     """Interpret a filename as either a filesystem location or as a package
     resource.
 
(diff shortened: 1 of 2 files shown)


pctablet505:
Withdrawing this one — going to step back from this issue for now rather than take up more review time. Apologies for the noise.

PR #1836, opened 2026-07-24
Fix spurious DROP TABLE in stamp --sql to base

### LINKED ISSUE #1822: Stamp command with "base" emits a DROP if --sql mode is present and no head revisions are present

Gooh456:
Fixes #1822.

`alembic stamp --sql base` emits a bare `DROP TABLE alembic_version` with no matching `CREATE TABLE` anywhere in the output. Real (non-sql) stamp never drops the version table, it only ever gets created, never dropped.

Root cause is in `MigrationContext.run_migrations()` (alembic/runtime/migration.py). At the end of the method there's an unconditional `if self.as_sql and not head_maintainer.heads: self._version.drop(...)`. This fires whenever an offline SQL script ends with zero heads, whether that's from stamp or from downgrade/upgrade steps that walk down to base. zzzeek confirmed on the issue this was a coding mistake from way back, added for --sql mode without checking that non-sql mode never does this.

Fix just removes that block so --sql output matches what actually happens against a real database.

This also changes `downgrade --sql <rev>:base` output, since it shares the same code path and had the same bug (there was an existing test, `test_version_to_none`, that was actually asserting the buggy DROP was present). Updated that test to assert no DROP, and added `test_sql_stamp_to_base_no_drop` for the stamp scenario from the issue.

Ran the full `tests/test_command.py` (116 tests) plus `test_version_traversal.py`, `test_environment.py`, `test_offline_environment.py` locally, all green.

### DIFF
--- a/alembic/runtime/migration.py
+++ b/alembic/runtime/migration.py
@@ -639,10 +639,6 @@ class MigrationContext:
                         run_args=kw,
                     )
 
-        if self.as_sql and not head_maintainer.heads:
-            assert self.connection is not None
-            self._version.drop(self.connection)
-
     def _in_connection_transaction(self) -> bool:
         try:
             meth = self.connection.in_transaction  # type: ignore[union-attr]
--- a/docs/build/unreleased/1822.rst
+++ b/docs/build/unreleased/1822.rst
@@ -1,0 +1,10 @@
+.. change::
+    :tags: bug, commands
+    :tickets: 1822
+
+    Fixed bug where the ``stamp`` command in ``--sql`` mode would emit a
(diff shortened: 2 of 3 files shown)


zzzeek:
dupe of #1823

Gooh456:
ah, missed #1823. same fix basically, no worries, all good closing this one.

Issue #681, opened 2020-04-08
Improve template logging and cookbook


CaselIT:
While using alembic I came across two areas that I think could be improved:

- in the template `evn.py` before setting the log config we should check if it's not already setup.
  I think that we can do it by checking if the root logger has an handler configured.
  This is useful if alembic is used programmatically, since it avoids resetting the program logging config.
  I guess that an option may comment may be enough for this, if the update in the template is too much
- Add a cookbook recipe on how to add a non nullable column without a server default value

CaselIT:
Regarding the second point, there are several options:
- add the column with a server_default, then remove it
- add the column as nullable, fill it with data using an update, then turn on nullable

Surely some other option I did not think of.

Is there a preferable way or both could/should be added? The second one is more powerful since we can also use conditions/case when satting the value

zzzeek:
people that want to add non-null columns often want to create a whole new table and then copy the old one over and rename.   this is because the migrations can lock the table and bring down a live site under load for a large set of data.

generally though for small things number 2 is likely more common, as removing server defaults can be a little weird for some backends.

zzzeek:
i think the important thing to note is that adding non-nullable cols is difficult to make work for an application that is running as the migrations go on.

zzzeek:
for the logging thing I dont know how logging would be set up already unless they are using the command environment progammatically, is that your use case?   how often does this get in your way that you notice logging is being configured twice when first setting up a project and you have to modify env.py (and if you are using programmatic config how are you not having to modify env.py for other reasons anyway)?

CaselIT:
Regarding non nullable then I guess we should suggest number 2, and add a warning regarding running applications with the copy etc.

CaselIT:
> is that your use case?

yes that is my use case. Like when creating a new db from scratch and stamping it with the current version.
And yes, I had changed the env to make it use the correct engine configuration from the application

CaselIT:
Maybe a simple comment would suffice, so that when an user first inits a project and creates the env and looks at it to adapt it to his application, can think if the logging check should be added in his use case.

CaselIT:
> people that want to add non-null columns often want to create a whole new table and then copy the old one over and rename. this is because the migrations can lock the table and bring down a live site under load for a large set of data.

[...]

PR #1838, opened 2026-07-29
Add cookbook recipe for adding a non-nullable column

ArockiaRajamanickam:
Adds the cookbook recipe requested in the second bullet of #681: how to add a non-nullable column to a table that already has rows, without a server default.

The recipe follows what was agreed in the thread rather than inventing an approach:

- The main recipe is CaselIT's option 2 (add as nullable, backfill with an `UPDATE`, then `alter_column(nullable=False)`), which was the one both of you settled on.
- A `.. warning::` covers zzzeek's two points: the `UPDATE`/`ALTER` can lock the table and take a live site down on a large dataset, and adding a non-nullable column is hard to make work against an application that keeps running during the migration.
- The `server_default` alternative is mentioned second, with the note that dropping a server default behaves differently across backends.
- Per CaselIT's comment, the create-new-table-and-copy approach is deliberately *not* suggested; the section links to `:ref:`batch_migrations`` instead.

The backfill uses the lightweight `sa.table()` / `sa.column()` constructs rather than the application model, so the migration keeps working as the model changes.

Scope: this PR only covers the cookbook half of #681. The `env.py` logging half was declined by zzzeek in the same thread, so it is left alone, and I have not used `Fixes:` since the issue bundles both requests.

Verification:

- Ran both recipes through a real `MigrationContext` against SQLite 3.53 and PostgreSQL 16: confirmed the naive one-step `add_column(nullable=False)` fails on a populated table, that the three-step recipe leaves an enforced `NOT NULL` column, that a subsequent `NULL` insert is rejected, that the downgrade drops the column, and that the `server_default` variant leaves no residual default.
- `sphinx -b html` builds with no new warnings; the `:ref:` and `:meth:` targets and the two SQLAlchemy intersphinx links all resolve.
- Code blocks are black-clean at the project's 79 character line length.

One note on the SQLite cross-reference: the `seealso` says the `nullable=False` step *may* need batch mode on backends with limited `ALTER` support, rather than stating flatly that SQLite cannot do it. While testing, SQLite 3.53 accepted `ALTER TABLE ... ALTER COLUMN ... SET NOT NULL` directly, so the absolute phrasing would not have been accurate on current versions. Happy to reword if you would rather it match the framing in `batch.rst`.

I used an AI assistant while drafting this; the wording and the testing above are mine and I have checked the content against the thread and the docs build.

### DIFF
--- a/docs/build/cookbook.rst
+++ b/docs/build/cookbook.rst
@@ -1732,6 +1732,19 @@ As the value is produced by an ``UPDATE`` statement, this approach also covers
 the case where the value depends on other columns, using a ``CASE`` expression
 or a correlated subquery.
 
+Step 3 emits ``ALTER TABLE ... ALTER COLUMN``, which most versions of SQLite do
+not accept, failing with ``near "ALTER": syntax error``. On SQLite, run that
+step inside a batch operation, which rebuilds the table instead::
+
+    # 3. on SQLite
+    with op.batch_alter_table("account") as batch_op:
+        batch_op.alter_column(
+            "status", existing_type=sa.String(50), nullable=False
+        )
+
+This form also works on the other backends, so it can be used unconditionally
+if the migration has to run everywhere.
+
 .. warning::
 
     The ``UPDATE`` in step 2 and the ``ALTER`` in step 3 may lock the table for
@@ -1774,9 +1787,9 @@ rather than solving it.
 
 .. seealso::
 
-    :ref:`batch_migrations` - on backends with limited ``ALTER`` support, such
-    as SQLite, the ``nullable=False`` step may need to run inside a batch
-    operation.
+    :ref:`batch_migrations` - the batch form of step 3 used above for SQLite,
+    and the other operations that backends with limited ``ALTER`` support
+    require it for.
 
 .. _custom_commandline:
 
--- a/docs/build/cookbook.rst
+++ b/docs/build/cookbook.rst
@@ -1686,6 +1686,98 @@ Writes are performed on both places, while the background script move all the re
 This technique is very challenging and time demanding, since it requires custom application logic to
 handle the intermediate states.
 
+.. _cookbook_non_nullable_column:
+
+Adding a Non-Nullable Column to an Existing Table
+=================================================
+
+Adding a column with ``nullable=False`` to a table that already contains rows
+is rejected by the database, as the existing rows would have no value for the
+new column and there is no default available to supply one::
+
+    # fails if "account" already has rows
+    op.add_column("account", sa.Column("status", sa.String(50), nullable=False))
+
+The usual way around this is to split the change into three steps inside the
+same migration: add the column as nullable, populate it for the rows that are
+already there, then apply the ``NOT NULL`` constraint::
+
+    from alembic import op
+    import sqlalchemy as sa
+
+
+    def upgrade():
+        # 1. add the column without the constraint
+        op.add_column("account", sa.Column("status", sa.String(50), nullable=True))
+
+        # 2. give the existing rows a value
+        account = sa.table("account", sa.column("status", sa.String))
+        op.execute(account.update().values(status="active"))
+
+        # 3. no NULL values remain, so the constraint can be applied
+        op.alter_column(
+            "account", "status", existing_type=sa.String(50), nullable=False
+        )
+
+
+    def downgrade():
+        op.drop_column("account", "status")
+
+The backfill in step 2 uses the lightweight
+:func:`~sqlalchemy.sql.expression.table` and
+:func:`~sqlalchemy.sql.expression.column` constructs instead of the
+application's model, so that the migration keeps working unchanged as the model
+evolves; a plain SQL string passed to :meth:`.Operations.execute` works as well.
+As the value is produced by an ``UPDATE`` statement, this approach also covers
+the case where the value depends on other columns, using a ``CASE`` expression
+or a correlated subquery.
+
+.. warning::
+
+    The ``UPDATE`` in step 2 and the ``ALTER`` in step 3 may lock the table for
+    the duration of the migration, which on a large table can take a live
+    application down. More generally, adding a non-nullable column is difficult
+    to make work against an application that keeps running while the migration
+    proceeds, since rows written by the old version of the application in
+    between steps 2 and 3 have no value for the new column. Migrating a live
+    application usually requires a deployment sequence designed for the
+    specific case, such as writing to both the old and the new column until the
+    backfill is complete.
+
+An alternative is to add the column with a ``server_default``, letting the
+database populate the existing rows, and then drop the default in a second
+step::
+
+    def upgrade():
+        op.add_column(
+            "account",
+            sa.Column(
+                "status",
+                sa.String(50),
+                nullable=False,
+                server_default="active",
+            ),
+        )
+        op.alter_column(
+            "account",
+            "status",
+            existing_type=sa.String(50),
+            existing_nullable=False,
+            server_default=None,
+        )
+
+This suits the case where one fixed value is correct for every existing row.
+Note that dropping a server default is handled differently from one backend to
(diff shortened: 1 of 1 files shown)


MohammedAlkindi (commented):
I executed the recipe rather than just reading it — docs that ship runnable steps deserve a runnable review. Windows 11, CPython 3.13.13, SQLAlchemy 2.x, in-memory SQLite, table pre-populated with two rows.

**What checks out, step by step:**
- The opening claim reproduces: `op.add_column(..., nullable=False)` on the populated table is rejected (`OperationalError`, "Cannot add a NOT NULL column with default value NULL" class of failure).
- Step 1 (nullable add) and step 2 (backfill through the lightweight `sa.table()`/`sa.column()` constructs) run exactly as written, and the backfill lands: both pre-existing rows read `status='active'`.
- After the constraint is applied, a `NULL` insert is rejected with `IntegrityError` — the end state is what the recipe promises.

**One real gap: step 3 as written fails on SQLite.**

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "ALTER": syntax error
[SQL: ALTER TABLE account ALTER COLUMN status SET NOT NULL]
```

SQLite has no `ALTER COLUMN`, so `op.alter_column(...)` emits SQL it cannot execute. The same three steps succeed if step 3 goes through batch mode — I verified this end-to-end (recipe completes, backfill correct, NULL insert rejected):

```python
with op.batch_alter_table("account") as batch_op:
    batch_op.alter_column("status", existing_type=sa.String(50), nullable=False)
```

Since batch mode is this project's own canonical answer to SQLite's ALTER limitations (and works unchanged on the other backends), a short note — or making batch the primary form with a line saying plain `alter_column` suffices on PostgreSQL/MySQL — would keep the cookbook honest for the one backend where the current text throws. Given how many people trial migrations against SQLite before pointing them at production, I'd guess this exact error is the first thing a fair fraction of readers would hit.

The lock-duration warning at the end is a good inclusion and matches what the operation actually does.

ArockiaRajamanickam:
Thank you for actually running it, that caught a real problem.

You are right, and I owe you a correction on the detail. When I wrote this I checked `ALTER TABLE ... ALTER COLUMN ... SET NOT NULL` against SQLite and found it worked, so I only softened the reference to batch mode into a `seealso` at the end of the section instead of putting it next to the step. Your report made me pin down why we disagree:

- SQLite **3.51.0**: `Error: in prepare, near "ALTER": syntax error` — exactly what you hit
- SQLite **3.53.0** and **3.53.1**: accepted, and genuinely applied. The schema becomes `status VARCHAR(50) NOT NULL`, a subsequent `NULL` insert raises `IntegrityError`, and running it against a table that already contains `NULL`s is refused rather than silently succeeding.

So the syntax arrived somewhere in 3.52/3.53, which is recent enough that whether the recipe works depends on the SQLite your Python happens to be linked against. My machine picks up 3.53, yours does not, and yours is the situation nearly every reader is in today. A `seealso` at the bottom of the section is the wrong place for something that decides whether the code above it runs at all.

Fixed in 4f553f8. The batch form now appears directly after step 3, with the actual error text so people can match it, plus a line noting the batch form works on the other backends too and can be used unconditionally.

I verified the batch version end to end the way you did: backfill lands on both rows, the resulting schema carries `NOT NULL`, a `NULL` insert is rejected with `IntegrityError`, and the pre-existing rows keep their data. Docs build succeeded with no new warnings, `cookbook.html` renders the new block, and the code block is black clean at 79 columns.

MohammedAlkindi:
Confirmed 4f553f8 resolves what I flagged, on a stock SQLite. Ran the updated recipe steps against SQLite 3.50.4 (alembic 1.19.1, SQLAlchemy 2.0.52, Python 3.13, Windows):

- Step 3 as originally written still fails there: `(sqlite3.OperationalError) near "ALTER": syntax error` — so the error text now quoted in the doc matches what a reader on a pre-3.52 SQLite sees, character for character.
- The new batch block works end to end: the rebuilt schema reads `status VARCHAR(50) NOT NULL`, both pre-existing rows keep their backfilled values, and a subsequent `NULL` insert raises `IntegrityError: NOT NULL constraint failed: account.status`.

3.50.4 is what a current CPython install links against, which supports the placement call: the batch form next to step 3, rather than in a `seealso`, is what most readers need today. Nothing left open from my review.

ArockiaRajamanickam:
Thanks for re-running it, and the 3.50.4 number is the useful part.

That is the piece I could not establish from here. My machines link 3.53, so I could only bracket the change between 3.51 (syntax error) and 3.53 (accepted and enforced). Your 3.50.4 being what a current CPython install ships confirms the practical point: for the foreseeable future almost every reader trying this against SQLite is on a version where the plain form throws, which is exactly why it needed to sit next to step 3 rather than in a `seealso` at the end.

Appreciate you checking the quoted error matches character for character. That was deliberate, so someone who hits it can search the page for what their traceback says.

MohammedAlkindi:
One thing that may explain the silence since: this PR looks like it never reached Gerrit, which is where alembic review actually happens. `sqla-tester` has not commented on it, and that comment is the signal that a change has been picked up.

Mine sat the same way. On #1847 I asked about it on 2026-08-20, CaselIT replied "Sorry i forgot to move it to gerrit", and `sqla-tester` posted the review link twenty seconds later.

So a one-line ping to a maintainer is probably all this needs. Nothing on your side looks wrong to me.

zzzeek:
hi -

doc PRs dont need to go to gerrit.     looking at this PR, it feels AI-written in how it has extra detail that isn't needed, while at the same time omitting other details that are arguably more important such as PostgreSQL lets you add a NOT NULL column directly if you give it a server default.    or it could be a computed column.   it's a whole subject area.

bigger picture I am hesitant to add recipes that dont really have anything to do with Alembic's API and are basically "how to use a database", because that implies we'd have "recipes" that describe all kinds of common database migration operations that are not at all specific to Alembic, and that seems like a thing that would be somewhere else.  cc @CaselIT  

anyway that's why #681 is just sitting there for six years until people started using LLMs to find open issues on github repos...

MohammedAlkindi:
Noted on Gerrit, and sorry for the misdirection. I generalised from #1847, which was a code change, and should have checked that doc PRs differ before pointing anyone that way.

ArockiaRajamanickam:
Thanks for looking, and that is a fair call. I will close this.

On scope you are right and it is your call to make. The recipe is mostly "how to add a NOT NULL column in SQL", with `alter_column` and `op.execute` incidental to it. If it went in, the argument for a dozen more like it gets easier, and none of them would be about Alembic's API either. That is a reasonable line to hold and I would rather not push you across it.

On the writing, the criticism lands. The `server_default` route is in there, but as a footnote after the three step version, which gets the emphasis backwards: on PostgreSQL adding the column with a server default is just the answer, not an alternative to the "usual" approach. Computed columns are not mentioned at all. You are right that it is a whole subject area, and a recipe that covers a third of it while going long on the parts it does cover is worse than no recipe.

I did use an AI assistant and said so at the bottom of the PR description when I opened it, so at least you were not guessing. That does not make the result better, and "extra detail that isn't needed while omitting more important things" is an accurate description of what came out.

On the last point, that is worth me taking seriously rather than brushing off. I did find #681 by searching open issues, and the honest lesson is that six years of silence on an issue is information in itself. I read it as neglected when it was more likely a scope question nobody had needed to answer out loud. I will weigh age the other way round in future.

#1839 is a different thing, a `TypeDecorator` wrapping `JSON` slipping past an `isinstance` check so every value in the column gets cast to `0` on a SQLite batch migration. That one is a bug fix with a regression test rather than a recipe, so I have left it open, but say the word if you would rather it went to Gerrit or nowhere.

@MohammedAlkindi thanks for digging into the Gerrit angle, and no harm done by the mixup.

CaselIT:
Tend to agree with mike here. What to to may be highly dependent on the context, so having a recepie for this may seem to implying that that is the one way you should go about adding a not nullable column.

PR #1840, opened 2026-08-03
Fix batch mode ignoring naming_convention for reflected constraints

### LINKED ISSUE #1834: add option to re-apply naming conventions for table renamed in batch mode
We thought this was a bug which we can see in discussion at #1833.  however, from the point of view of "batch migrations" as a workaround for the lack of ALTER, it's not a bug.

1. Batch migrations recreate a table with a new name, then drop the old table and rename the new table to the old one
2. it normally uses reflection to get the full structure of the existing table, including all of its constraints (and their names).  that is, the constraints that are already created in the database.
3. a SQLite bug that was fixed in SQLAlchemy 2.0.45 https://github.com/sqlalchemy/sqlalchemy/issues/12924 had the behavior such that SQLite constraints where the constraint name used MixedCase would not reflect the constraint name; we'd get None for the constraint name in that case
4. when batch reflection / table recreation gets None for a constraint name, it uses the configured naming convention (if any) to give the new constraint a name
5. user @mbergen had a schema in SQLite using a naming convention where constraints had MixedCase.
6. up through SQLAlchemy 2.0.44, @mbergen relied upon the behavior that a batch migration for SQLite where the table name changed would also rename their constraints for that new table name, so a constraint called "Old_Table_ck_1" would nicely become "New_Table_ck_1" (noting MixedCase names)
7. the bug fixed in 2.0.45 meant the operation would now successfully reflect the constraint's name of "Old_Table_ck_1" and leave it unchanged when creating a new table.
8. the behavior for constraint names with lowercase on SQLite as well as PG, MySQL was always that the constraint name was reflected, and applying a table rename and/or a batch recreate didnt change those names.
9. this behavior also matches what happens if we use ALTER TABLE.  If you have a table named "Old_Table" and your naming convention made a constraint "Old_Table_ck_1", and you do "ALTER TABLE 'Old_Table' RENAME TO 'New_Table'", that's not going to rename your "Old_Table_ck1" constraint - that old name stays.
10. this all means the current behavior plus the 2.0.45 sqlite fix for alembic is "correct", for a "correct" of "batch migrations are a substitute for ALTER TABLE".

So.   With all of the above establishing this was not a bug, the batch mode thing does give us a way to...rename a table and drop and recreate all its constraints!  so maybe it would be nice for there to be an option, "rename_constraints_with_convention=<some convention>" or something like that, so that you **can** actually have that behavior @mbergen had come to rely upon.

that's the issue here.   feature request

Karthick-dev-cart:
Fixes: #1834

### Description

`batch_alter_table(..., naming_convention=...)` reflects the existing table into a `MetaData(naming_convention=...)`, clearly intending for the convention to govern the recreated table's constraint names. But reflection always returns constraints with an explicit `.name` (whatever is actually stored in the database), and SQLAlchemy naming conventions only fill in names for constraints that don't already have one — so the convention silently had no effect on any reflected constraint.

This was most visible after `rename_table()`: batch mode would faithfully carry over reflected constraint names that still embed the *old* table name (e.g. `pk_OldTableName` after renaming to `TableName`), even though a `naming_convention` was explicitly passed to make names track the current table.

### Fix

In `ApplyBatchImpl._grab_table_elements()`, a reflected constraint whose type has an entry in the supplied `naming_convention` (pk/fk/uq/ck) now has its name reset to `None` and is re-attached to `self.table` — which already carries the correct, final table name and the convention — so SQLAlchemy's own naming-convention event recomputes the name against the right target immediately, before the constraint gets copied onto the temp-named table used internally during the "move and copy". (An earlier version of this fix deferred the regeneration to copy time instead, which "fixed" the stale-old-name bug but introduced a different one — baking `_alembic_tmp_<name>` into the constraint name. Caught via a test that isolates the no-rename case and fixed by regenerating against `self.table` up front.)

Constraint types the convention doesn't mention are left untouched, so this can't turn some other constraint into an unnamed one it wasn't asked to govern.

`_transfer_elements_to_new_table()` now also carries `naming_convention` onto the new table's `MetaData` (previously always plain `MetaData()`), which the reset-name mechanism above relies on.

### Checklist

- [x] A short code fix
- Issue: #1834 (includes a complete reproduction and root-cause diagnosis)
- `Fixes: #1834` included above
- Tests: `tests/test_batch.py::BatchRoundTripTest::test_naming_convention_applies_to_reflected_constraints` — reproduces the issue's exact scenario (rename + batch recreate with a naming_convention covering pk/fk), asserts the recreated table's PK and FK names track the *new* table name. Verified this test fails against unpatched `main` with the exact bug signature (`pk_oldtablename != pk_tablename`).

### Test plan

- New regression test passes; fails on unpatched `main` with the issue's exact symptom.
- Full suite: `pytest tests/` → 1799 passed, 132 skipped (pre-existing, dialect-specific), 0 failed.
- `black --check` / `flake8` clean on both changed files.
- Changelog fragment added at `docs/build/unreleased/1834.rst`.

### DIFF
--- a/alembic/operations/batch.py
+++ b/alembic/operations/batch.py
@@ -21,6 +21,7 @@ from sqlalchemy import schema as sql_schema
 from sqlalchemy import select
 from sqlalchemy import Table
 from sqlalchemy import types as sqltypes
+from sqlalchemy import UniqueConstraint
 from sqlalchemy.sql.schema import SchemaEventTarget
 from sqlalchemy.util import OrderedDict
 from sqlalchemy.util import topological
@@ -155,6 +156,7 @@ class BatchOperationsImpl:
                     self.table_kwargs,
                     reflected,
                     partial_reordering=self.partial_reordering,
+                    naming_convention=self.naming_convention,
                 )
                 for opname, arg, kw in self.batch:
                     fn = getattr(batch_impl, opname)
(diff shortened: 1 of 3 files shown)


zzzeek:
we have a WIP for this at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6832/ and we're also still deciding exactly how we want to implement the option for this behavior (no longer a bug)

PR #1843, opened 2026-08-08
Fix: preserve quoted_name(quote=...) semantics when rendering schema in autogenerate

### LINKED ISSUE #1526: Migration autogeneration should respect schemas defined using quoted_name
**Describe the bug**
When defining a table, schemas defined using quoted_name should be persisted on the migration generated code.

**Expected behavior**
The schema present in the migration file should be defined using quoted_name too

**To Reproduce**

```py
from sqlalchemy.orm import DeclarativeBase
from sqlalchemy import quoted_name

class Base(DeclarativeBase):
    pass

class Example(Base):
    __table_args__ = {'schema': quoted_name('database.schema', quote=False)}
    example_field = Column(Integer, primary_key=True)
```

**Actual behaviour**

```py
    op.create_table('example',
    sa.Column('example_field', sa.Integer()),
    sa.PrimaryKeyConstraint('account_id'),
    schema='datahub.kamino'
    )

```

**Expected behaviour**

```py
    op.create_table('example',
    sa.Column('example_field', sa.Integer()),
    sa.PrimaryKeyConstraint('account_id'),
    schema=quoted_name('database.schema', quote=False)
    )

```

**Versions.**
 - OS: Mac
 - Python:3.12
 - Alembic: 1.13.2
 - SQLAlchemy: 2.0.32
 - Database: Databricks

**Additional context**
Im using the databricks sql connector 3.3.0

**Have a nice day!**

pranjalm37:
## Summary

Fixes #1526.

When a table/index/constraint's schema is set using an explicit
`quoted_name(name, quote=True)` or `quoted_name(name, quote=False)`, the
autogenerate render code (`_ident()` in `alembic/autogenerate/render.py`)
collapsed the value down to a plain string before formatting it with `%r`,
so the caller's explicit quoting preference was silently dropped from the
generated migration file.

## Fix

Adds `_render_schema_ident()`, used at every `schema=` render site
(create/drop table, index, column, check/unique constraint, drop_constraint,
alter_column, and the `batch_alter_table` header):

- If the schema is a `quoted_name` with `quote` explicitly set (`True`/`False`),
  render it as `quoted_name('name', True/False)` and add the necessary import
  to the generated migration file.
- Otherwise, falls back to `repr()` of the original value, preserving prior
  behavior — including for objects with a custom `__repr__` (there's an
  existing test, `test_drop_unique_constraint_schema_reprobj`, covering that
  case, which still passes).

## Test plan

- [x] Added `test_render_table_w_quoted_name_schema`, matching the repro from
      #1526, asserting both the rendered code and that the `quoted_name` import
      is added.
- [x] Full test suite passes locally: `pytest tests/ --ignore=tests/test_post_write.py --ignore=tests/test_mssql.py --ignore=tests/test_oracle.py --ignore=tests/test_mysql.py --ignore=tests/test_postgresql.py` → 1562 passed, 133 skipped (DB-specific suites skipped, no local DB available; `test_post_write.py` failures are pre-existing/unrelated — they fail identically on `main` without this change, due to `black` not being installed in the hook environment).
- [x] `black --check` and `flake8` pass on both changed files.
- [x] Added changelog fragment at `docs/build/unreleased/1526.rst`.

### DIFF
--- a/alembic/autogenerate/render.py
+++ b/alembic/autogenerate/render.py
@@ -142,8 +142,11 @@ def _render_modify_table(
         if render_as_batch:
             with autogen_context._within_batch():
                 lines.append(
-                    "with op.batch_alter_table(%r, schema=%r) as batch_op:"
-                    % (op.table_name, op.schema)
+                    "with op.batch_alter_table(%r, schema=%s) as batch_op:"
+                    % (
+                        op.table_name,
+                        _render_schema_ident(autogen_context, op.schema),
+                    )
                 )
                 for t_op in op.ops:
                     t_lines = render_op(autogen_context, t_op)
@@ -258,7 +261,9 @@ def _add_table(autogen_context: AutogenContext, op: ops.CreateTableOp) -> str:
         "args": args_str,
     }
     if op.schema:
-        text += ",\nschema=%r" % _ident(op.schema)
+        text += ",\nschema=%s" % _render_schema_ident(
+            autogen_context, op.schema
+        )
 
     comment = table.comment
     if comment:
@@ -289,7 +294,9 @@ def _drop_table(autogen_context: AutogenContext, op: ops.DropTableOp) -> str:
         "tname": _ident(op.table_name),
     }
     if op.schema:
-        text += ", schema=%r" % _ident(op.schema)
+        text += ", schema=%s" % _render_schema_ident(
+            autogen_context, op.schema
+        )
 
     if op.if_exists is not None:
         text += ", if_exists=%r" % bool(op.if_exists)
@@ -338,7 +345,10 @@ def _add_index(autogen_context: AutogenContext, op: ops.CreateIndexOp) -> str:
         ),
         "unique": index.unique or False,
         "schema": (
-            (", schema=%r" % _ident(index.table.schema))
+            (
+                ", schema=%s"
+                % _render_schema_ident(autogen_context, index.table.schema)
+            )
             if index.table.schema
             else ""
         ),
@@ -367,7 +377,11 @@ def _drop_index(autogen_context: AutogenContext, op: ops.DropIndexOp) -> str:
         "prefix": _alembic_autogenerate_prefix(autogen_context),
         "name": _render_gen_name(autogen_context, op.index_name),
         "table_name": _ident(op.table_name),
-        "schema": ((", schema=%r" % _ident(op.schema)) if op.schema else ""),
+        "schema": (
+            (", schema=%s" % _render_schema_ident(autogen_context, op.schema))
+            if op.schema
+            else ""
+        ),
         "kwargs": ", " + ", ".join(opts) if opts else "",
     }
     return text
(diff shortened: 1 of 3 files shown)


MohammedAlkindi (commented):
Ran this on Windows 11 / Python 3.13.13, since alembic's CI matrix is `ubuntu-22.04` / `ubuntu-latest` only.

**No new failures.** Full suite on this branch:

```
$ python -m pytest tests/ -q
5 failed, 1814 passed, 132 skipped in 39.87s
```

The 5 are all `ScriptNamingTest` timezone tests (`test_custom_tz`, `test_custom_tz_lowercase`, `test_custom_tz_utc`, `test_custom_tzdata_tz`, `test_generates_a_date`) failing on `zoneinfo` with no system tzdata, which is a Windows-only environment gap. They fail identically on a clean `upstream/main` checkout on this machine — same count, same five names — so the delta from this PR is zero.

**The new test has teeth.** Reverting only the source file and keeping the test:

```
$ git checkout upstream/main -- alembic/autogenerate/render.py
$ python -m pytest tests/test_autogen_render.py -q
FAILED tests/test_autogen_render.py::AutogenRenderTest::test_render_table_w_quoted_name_schema
AssertionError:
  "...schema='database.schema')"
  != "...schema=quoted_name('database.schema', False))"
1 failed, 178 passed, 2 skipped
```

and with `render.py` restored, `179 passed, 2 skipped`. So it fails before and passes after rather than only covering the happy path — the rendered output genuinely loses the `quote=False` marker without the change, which is the reported bug.

Not checked: PostgreSQL/MySQL/Oracle behaviour (I ran the default sqlite configuration), and I did not review the `render.py` changes for design intent — this is a platform and regression-coverage report, not a design review.

MohammedAlkindi (commented):
Checked this out and ran it on Windows 11 (10.0.26200), Python 3.13.13, against `main` at `c116cbc0`.

**Test evidence:**

| `pytest tests/test_autogen_render.py` | passed | skipped |
|---|---|---|
| `main` | 178 | 2 |
| this PR | **179** | 2 |

The +1 is the PR's own test, and no existing test changes state.

**Fail-before confirmed**, which is the part worth having: with `alembic/autogenerate/render.py` reverted to `main` and the new test kept, `test_render_table_w_quoted_name_schema` **fails**. So the test genuinely discriminates the fix rather than just documenting current behaviour.

**The diagnosis looks right to me.** `quoted_name` is a `str` subclass, so `%r` collapses it to a plain string literal and the `quote=` flag simply disappears from the generated migration — the rendered file then re-renders with default quoting rules, which is exactly the #1526 symptom. Routing through `_render_schema_ident` and emitting the explicit wrapper preserves the caller's intent.

**Coverage check:** I grepped `render.py` for any remaining schema render that still goes through `%r`, and found none — `_add_table`, `_drop_table`, `_render_modify_table` and the index path all go through the new helper now. The change looks complete rather than partial.

**One thing I checked rather than flagged.** `_render_schema_ident` returns `str | None`, and `_render_modify_table` interpolates it with `%s` *without* an `if op.schema:` guard, unlike `_add_table` and `_drop_table`. That looked like it might emit a literal `schema=None`, but it's fine and unchanged: `%r` of `None` and `%s` of `None` both produce `None`, so a schema-less batch op renders `schema=None` exactly as it did before. Noting it only so the asymmetry between guarded and unguarded call sites doesn't read as an oversight later.

I ran only `test_autogen_render.py` rather than the full suite, since that is the file this touches — worth saying explicitly so the numbers above aren't read as a whole-suite result.

---

*Disclosure: I used an AI assistant while reviewing. The before/after runs, the fail-before check, and the `render.py` grep were executed by me on the machine described.*

Issue #1845, opened 2026-08-08
batch_alter_table naming_convention re-applies ck convention to reflected constraint names


zzzeek:
**Describe the bug**

`batch_alter_table(naming_convention=...)` reflects the existing table into a
`MetaData` that carries the convention. Reflected constraints already have names,
but a `ck` convention containing `%(constraint_name)s` is re-applied to them, so
the existing name gets prefixed a second time.

Only conventions using the `%(constraint_name)s` token are affected; `uq`/`fk`/`pk`
conventions don't include it, so reflected names pass through unchanged there.

**To Reproduce**

```py
from alembic.migration import MigrationContext
from alembic.operations import Operations
from sqlalchemy import CheckConstraint, Column, Integer, MetaData, Table, create_engine

naming = {"ck": "ck_%(table_name)s_%(constraint_name)s"}

m = MetaData(naming_convention=naming)
Table(
    "user",
    m,
    Column("id", Integer, primary_key=True),
    Column("age", Integer),
    CheckConstraint("age > 0", name="positive_age"),
)

e = create_engine("sqlite://")
m.create_all(e)
with e.begin() as conn:
    print("before:", conn.exec_driver_sql(
        "select sql from sqlite_master where name='user'").scalar())

    op = Operations(MigrationContext.configure(conn))
    with op.batch_alter_table(
        "user", naming_convention=naming, recreate="always"
    ) as batch_op:
        batch_op.add_column(Column("y", Integer))

    print("after: ", conn.exec_driver_sql(
        "select sql from sqlite_master where name='user'").scalar())
```

**Expected behavior**

The reflected constraint keeps its existing name; the convention should only supply
names for constraints that don't have one.

**Actual behavior**

```
before: CREATE TABLE user (
	id INTEGER NOT NULL,
	age INTEGER,
	PRIMARY KEY (id),
	CONSTRAINT ck_user_positive_age CHECK (age > 0)
)
after:  CREATE TABLE "user" (
	id INTEGER NOT NULL,
	age INTEGER,
	y INTEGER,
	PRIMARY KEY (id),
	CONSTRAINT ck_user_ck_user_positive_age CHECK (age > 0)
)
```

The prefix accumulates on every recreate.

**Additional context**

The renaming happens at reflection time in `BatchOperationsImpl.flush()`, where the
table is reflected into `MetaData(naming_convention=self.naming_convention)`.
Discovered while investigating #1768; related to #1844.

**Versions**
[...]

PR #1848, opened 2026-08-10
Fix: In ApplyBatchImpl.flush() (alembic/operations/batch.py), when...

M001N:
## Summary
After reflecting the existing table into the naming_convention-bearing MetaData (m1) as before (this reflection is still needed so genuinely unnamed constraints -- e.g. some unnamed FKs on SQLite -- still get a convention-generated name, matching existing test_drop_foreign_key behavior), the table is reflected a second time into a plain MetaData() with no naming_convention. This second reflection yields each constraint's true, un-mangled database name. Constraints from the first (convention-bearing) reflection are then matched to their plain-reflected counterpart by (type, column names) and, when the plain-reflected constraint already had a real name, that true name is restored onto the convention-reflected constraint, undoing any double-prefixing. Constraints with no name in the plain reflection are left untouched so the convention can still generate a name for them.

## Problem
sqlalchemy/alembic issue reference: sqlalchemy/alembic#1845

## Root Cause
In ApplyBatchImpl.flush() (alembic/operations/batch.py), when should_recreate is True the existing table is reflected via Table(..., autoload_with=...) into a MetaData(naming_convention=self.naming_convention). SQLAlchemy's naming convention machinery re-applies the %(constraint_name)s token substitution to a constraint's name even when it was reflected and already has a real, persisted name -- it does not distinguish 'freshly reflected, already named' from 'newly created, needs a generated name'. This is only user-visible for convention templates that include %(constraint_name)s (e.g. 'ck'); uq/fk/pk templates in the repo's own tests don't use that token so they weren't previously affected.

## Testing
PASS - full tests/test_batch.py suite: 120 passed, 7 skipped (skips are pre-existing, backend-specific), including the two new regression tests and the pre-existing naming_convention test_drop_foreign_key (fk convention with %(column_0_name)s etc., unaffected/still passing).

## Related Issue
sqlalchemy/alembic#1845

### DIFF
--- a/alembic/operations/batch.py
+++ b/alembic/operations/batch.py
@@ -143,6 +143,58 @@ class BatchOperationsImpl:
                     )
                     reflected = True
 
+                    if self.naming_convention:
+                        # reflected constraints already have a real,
+                        # persisted name.  however, a naming convention
+                        # such as {"ck": "ck_%(table_name)s_%(constraint_name)s"}  # noqa: E501
+                        # gets re-applied to that name during the above
+                        # reflection since the constraint's name is not
+                        # distinguished from a freshly-assigned one,
+                        # causing it to be prefixed a second time.
+                        # reflect again into a plain MetaData() (no
+                        # naming convention) to determine each
+                        # constraint's real, un-mangled name, and
+                        # restore it.  constraints that are genuinely
+                        # unnamed in the database are left alone so the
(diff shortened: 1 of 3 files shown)


zzzeek:
sorry I have to ask you to stop submitting PRs.   these issues have patches in internal review already

Issue #1844, opened 2026-08-08
Batch recreate loses naming convention for Boolean/Enum CHECK constraints


zzzeek:
**Describe the bug**

In batch "recreate" mode, a CHECK constraint generated by a `SchemaType`
(`Boolean`/`Enum` with `create_constraint=True`) loses its naming convention name.
The constraint is regenerated from the column's type onto the temporary table, which
is built on a plain `MetaData()` with no naming convention, so it comes out with the
raw name from the type rather than the convention-applied name.

This is specific to passing a ``Table`` object as ``copy_from``, which is what
carries a live ``SchemaType`` into the recreate.  In the reflected case the
constraint comes back from reflection as an ordinary named CHECK and is
transferred verbatim, so it is unaffected.

No column operation on the affected column is needed; any batch operation that
triggers a recreate is enough.

This is distinct from the limitation documented at
https://alembic.sqlalchemy.org/en/latest/batch.html#batch-check-constraints,
which concerns *unnamed* CHECK constraints being omitted from the recreate.
Here the type is named, so the constraint does participate in the recreate --
just under a different name than it had.

**To Reproduce**

```py
from alembic.migration import MigrationContext
from alembic.operations import Operations
from sqlalchemy import Boolean, Column, Integer, MetaData, Table, create_engine

naming = {"ck": "ck_%(table_name)s_%(constraint_name)s"}

m = MetaData(naming_convention=naming)
user = Table(
    "user",
    m,
    Column("id", Integer, primary_key=True),
    Column("is_active", Boolean(create_constraint=True, name="is_active")),
)

e = create_engine("sqlite://")
m.create_all(e)

with e.begin() as conn:
    op = Operations(MigrationContext.configure(conn))
    with op.batch_alter_table("user", copy_from=user, recreate="always") as batch_op:
        batch_op.add_column(Column("y", Integer))

    print(conn.exec_driver_sql(
        "select sql from sqlite_master where name='user'").scalar())
```

**Expected behavior**

The constraint keeps the name it had, matching `create_all()`:

```sql
CONSTRAINT ck_user_is_active CHECK (is_active IN (0, 1))
```

**Actual behavior**

```
CREATE TABLE "user" (
	id INTEGER NOT NULL,
	is_active BOOLEAN,
	y INTEGER,
	PRIMARY KEY (id),
	CONSTRAINT is_active CHECK (is_active IN (0, 1))
)
```

[...]

PR #1849, opened 2026-08-10
Fix: In BatchOperationsImpl.flush() (alembic/operations/batch.py), the...

M001N:
## Summary
Added a `naming_convention` parameter to ApplyBatchImpl.__init__, stored as self.naming_convention. Updated _transfer_elements_to_new_table() to build the temp table's MetaData with `MetaData(naming_convention=self.naming_convention)` when set, matching the pattern already used for m1 in flush(). Updated BatchOperationsImpl.flush() to pass self.naming_convention through when constructing ApplyBatchImpl.

## Problem
sqlalchemy/alembic issue reference: sqlalchemy/alembic#1844

## Root Cause
In BatchOperationsImpl.flush() (alembic/operations/batch.py), the copy_from branch reuses the caller's Table object verbatim (its own MetaData, no naming_convention) as expected. However, the actual constraint materialization for SchemaType-generated CHECK constraints happens later in ApplyBatchImpl._transfer_elements_to_new_table(), which builds the temp table on `m = MetaData()` -- a brand new MetaData with no naming_convention, regardless of what convention was passed to batch_alter_table(). Boolean/Enum(create_constraint=True) columns regenerate their CheckConstraint when attached to this temp table, and that regeneration resolves its name via the temp table's MetaData.naming_convention -- which was always empty. This affects both the copy_from and reflected paths (both funnel through the same temp-table construction), but is specifically what the issue describes for copy_from.

## Testing
PASS - full tests/test_batch.py suite: 119 passed, 7 skipped (unrelated DB-specific skips), 0 failures, including the new regression test and all pre-existing naming-convention/CopyFromTest tests.

## Related Issue
sqlalchemy/alembic#1844

### DIFF
--- a/alembic/operations/batch.py
+++ b/alembic/operations/batch.py
@@ -149,6 +149,7 @@ class BatchOperationsImpl:
                     self.table_args,
                     self.table_kwargs,
                     reflected,
+                    naming_convention=self.naming_convention,
                     partial_reordering=self.partial_reordering,
                 )
                 for opname, arg, kw in self.batch:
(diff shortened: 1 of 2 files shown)


zzzeek:
sorry I have to ask you to stop submitting PRs.   these issues have patches in internal review already

PR #1855, opened 2026-08-12
Fix BatchOperations.create_index typing

### LINKED ISSUE #1853: Bug: BatchOperations.create_index has incorrect signature
### Discussed in https://github.com/sqlalchemy/alembic/discussions/1850

<sup>Originally posted by **Annonymus-v02** August 11, 2026</sup>
## Describe the bug
The signature (according to the docs and the API in the generated section of base.py) of BatchOperations.create_index is 
```python
def create_index(self, index_name: str, columns: list[str], **kw: Any) -> None
```
The signature of Operations.create_index is 
```python
def create_index(self, index_name: str | None, table_name: str, columns: Sequence[str | TextClause | ColumnElement[Any]], *, schema: str | None = None, unique: bool = False, if_not_exists: bool | None = None, **kw: Any) -> None
```
Having taken a very brief look at where these functions come from I believe the signatures must be the same (aside from the missing table_name) as they are backed by the same implementation, but my (possibly — and even somewhat likely — incorrect) beliefs aside, `alembic revision --autogenerate` does not respect BatchOperations' version of the signature, generating code like 
```python
with op.batch_alter_table('table_name', schema=None) as batch_op:
    batch_op.create_index(batch_op.f('idx_name'), [sa.literal_column('"col1_name" DESC'), sa.literal_column('"col2_name" DESC')], unique=False)
```
which is not valid for BatchOperations' version, which declares `columns` to be a `list[str]`.
Given that this code actually runs (generates a `CREATE INDEX` statement using `alembic upgrade --sql`), I believe this is a bug in the signature, not in `--autogenerate`'s handling of batch operations.

`create_index` is the function that triggered our type checker, making us aware of this issue. I have not verified whether it is the only affected function, or there are other mismatches between the signature in `Operations` and `BatchOperations`.

## Expected behavior
The signature of `BatchOperations.create_index` should reflect the parameters it actually expects, or alternatively, `--autogenerate` should not generate code that violates the current signature.

## To Reproduce
Run `alembic upgrade b23c5a130bc8 --sql` with the following migration and observe that no error occurs and the resulting sql file contains a CREATE INDEX statement.
Run your preferred type checker and observe the type violation reported on line 16, or read [the documentation](https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.BatchOperations.create_index) and notice that it defines the `columns` parameter as a `list[str]`, which is not the case in the code below.
```py
from typing import Sequence, Union

from alembic import op
import sqlalchemy as sa

# revision identifiers, used by Alembic.
revision: str = 'b23c5a130bc8'
down_revision: Union[str, Sequence[str], None] = None
branch_labels: Union[str, Sequence[str], None] = None
depends_on: Union[str, Sequence[str], None] = None


def upgrade(engine_name: str) -> None:
    """Upgrade schema."""
    with op.batch_alter_table('prm_contacts', schema=None) as batch_op:
        batch_op.create_index(batch_op.f('idx_test_idx_delete_me'), [sa.literal_column('"NAME" DESC'), sa.literal_column('"PHONE" DESC')], unique=False)


def downgrade(engine_name: str) -> None:
    """Downgrade schema."""
    with op.batch_alter_table('prm_contacts', schema=None) as batch_op:
        batch_op.drop_index(batch_op.f('idx_test_idx_delete_me'))
```

## Error
In this example using `ty`
```
error[invalid-argument-type]: Argument to bound method `BatchOperations.create_index` is incorrect
  --> alembic/versions/2026_08_11-b23c5a130bc8_test_revision.py:16:69
   |
16 | …dx_test_idx_delete_me'), [sa.literal_column('"NAME" DESC'), sa.literal_column('"PHONE" DESC')], unique=False)
   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Expected `list[str]`, found `list[str | ColumnClause[Unknown]]`
info: element `ColumnClause[Unknown]` of union `str | ColumnClause[Unknown]` is not assignable to `str`
info: Method defined here
    --> .venv/lib/python3.14/site-packages/alembic/operations/base.py:1871:13
     |
1871 |         def create_index(
     |             ^^^^^^^^^^^^
1872 |             self, index_name: str, columns: List[str], **kw: Any
     |                                    ------------------ Parameter declared here
```

## Versions
 - OS: MacOS Tahoe 26.5.1
 - Python: 3.14.6
 - Alembic: 1.19.1 — also observed on version 1.18.4
 - SQLAlchemy: 2.0.46
 - Database: irrelevant — behavior can be observed running offline (`--sql`)[^1], but we originally discovered it on a migration for an oracle DB
 - DBAPI: not sure what this is asking for (db driver version?), but I believe it is irrelevant, as above.
 
 [^1]: or not even running at all just looking at the `--autogenerate`d code and the documentation.

## Additional context
I am writing this bug report as a discussion since the [reporting bugs](https://www.sqlalchemy.org/participate.html#bugs) guidance asks to do so instead of opening an issue, but none of the categories of discussions look right... I chose the one that seems least inappropriate and marked the actual nature in the title. Please let me know if that was not the right choice.</div>

Boulea7:
Fixes: #1853

### Description

Update `BatchOperations.create_index()` so its generated API accepts the same column expression types as `Operations.create_index()`. This removes false type-checker errors for batch migrations that use expressions such as `literal_column()`.

The `BatchOperations` API was regenerated with `tools/write_pyi.py`. Generated-stub consistency, the repository mypy check, and the SQLite test suite pass.

### DIFF
--- a/alembic/operations/base.py
+++ b/alembic/operations/base.py
@@ -1857,7 +1857,10 @@ class BatchOperations(AbstractOperations):
             ...
 
         def create_index(
-            self, index_name: str, columns: list[str], **kw: Any
+            self,
+            index_name: str,
+            columns: Sequence[str | TextClause | ColumnElement[Any]],
+            **kw: Any,
         ) -> None:
             """Issue a "create index" instruction using the
             current batch migration context.
--- a/alembic/operations/ops.py
+++ b/alembic/operations/ops.py
@@ -1008,7 +1008,7 @@ class CreateIndexOp(MigrateOperation):
         cls,
         operations: BatchOperations,
         index_name: str,
-        columns: list[str],
+        columns: Sequence[str | TextClause | ColumnElement[Any]],
         **kw: Any,
     ) -> None:
         """Issue a "create index" instruction using the


sqla-tester:
Thanks! Issue #1853 is now marked **code review in progress** and no longer **open for pull requests**, so this pull request holds the review for it and another one won't land on top of your work. If this pull request is abandoned, a maintainer can put **open for pull requests** back on #1853 to reopen it to others.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 36a42d5934050181dfd28441567c4b94a9a4c0c1 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change 36a42d5934050181dfd28441567c4b94a9a4c0c1: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6897

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6897 has been **merged**. Congratulations! :)

PR #1858, opened 2026-08-19
Warn when batch recreate skips unnamed reflected CHECK constraints

### LINKED ISSUE #1846: Batch recreate silently drops unnamed reflected CHECK constraints without warning
**Describe the bug**

Batch "recreate" silently discards any unnamed CHECK constraint on a reflected
table.  The migration reports success and the constraint is simply gone from the
recreated table.

This is documented behavior -- see
https://alembic.sqlalchemy.org/en/latest/batch.html#batch-check-constraints -- and the
documented escapes (name the type, or restate the constraint via ``table_args``)
work.  The problem is that nothing tells the user it happened.  Since the outcome
is the loss of a constraint rather than a failure, there's no signal at the time
and no signal afterwards.

This affects any unnamed CHECK constraint, not only the ones generated by
``Boolean``/``Enum``; a hand-written ``CheckConstraint("x > 5")`` with no name is
discarded the same way.

**To Reproduce**

```py
from alembic.migration import MigrationContext
from alembic.operations import Operations
from sqlalchemy import (Boolean, CheckConstraint, Column, Integer, MetaData,
                        Table, create_engine)

m = MetaData()          # no naming convention involved
Table(
    "user",
    m,
    Column("val", Boolean(create_constraint=True)),   # type has no .name
    Column("x", Integer),
    CheckConstraint("x > 5"),                         # hand-written, unnamed
)

e = create_engine("sqlite://")
m.create_all(e)

with e.begin() as conn:
    def show(label):
        sql = conn.exec_driver_sql(
            "select sql from sqlite_master where name='user'").scalar()
        checks = [l.strip().rstrip(",") for l in sql.splitlines() if "CHECK" in l]
        print(label, checks or "*** NONE ***")

    show("before:")
    op = Operations(MigrationContext.configure(conn))
    with op.batch_alter_table("user", recreate="always") as batch_op:
        batch_op.add_column(Column("y", Integer))
    show("after: ")
```

**Expected behavior**

A warning stating that the constraint is being omitted from the table recreate,
pointing at the two documented ways to keep it.

**Actual behavior**

```
before: ['CHECK (x > 5)', 'CHECK (val IN (0, 1))']
after:  *** NONE ***
```

No warning, no error.

**Additional context**

The constraints are dropped by the ``self.reflected and isinstance(const,
CheckConstraint) and not const.name`` branch of
``ApplyBatchImpl._grab_table_elements()``, which carries a longstanding TODO
noting there's no way to determine ``_is_type_bound()`` for a reflected
constraint.  That reasoning still holds -- this issue asks only for a warning at
that point, not for the constraints to be retained.

That branch currently has no test coverage; instrumenting it and running the full
test suite produces zero hits.

Noticed while working on #1768 / #1844 / #1845.

**Versions**
 - Alembic: 1.19.1 (main)
 - SQLAlchemy: 2.1
 - Database: SQLite

official-burak:
Fixes #1846.

Batch recreate still cannot round-trip unnamed reflected CHECK constraints, because they cannot be distinguished from type-bound Boolean/Enum checks. That skip used to be silent, so a constraint could disappear with no signal.

Emit a UserWarning that names the table (and the constraint SQL when available) and points at naming the constraint or restating it via `table_args`.

### DIFF
--- a/alembic/operations/batch.py
+++ b/alembic/operations/batch.py
@@ -20,6 +20,7 @@ from sqlalchemy.sql.schema import SchemaEventTarget
 from sqlalchemy.util import OrderedDict
 from sqlalchemy.util import topological
 
+from .. import util
 from ..util import exc
 from ..util.sqla_compat import _columns_for_constraint
 from ..util.sqla_compat import _copy
@@ -263,10 +264,17 @@ class ApplyBatchImpl:
                 and isinstance(const, CheckConstraint)
                 and not const.name
             ):
-                # TODO: we are skipping unnamed reflected CheckConstraint
-                # because
-                # we have no way to determine _is_type_bound() for these.
-                pass
+                # Unnamed reflected CHECKs cannot be distinguished from
+                # type-bound constraints, so they are not recreated. Warn
+                # rather than dropping them silently.
+                sqltext = getattr(const, "sqltext", None)
+                detail = f" ({sqltext})" if sqltext is not None else ""
+                util.warn(
+                    "Skipping unnamed CHECK constraint on reflected "
+                    "table %r during batch recreate%s. Name the "
+                    "constraint, or restate it via table_args, if it "
+                    "should be preserved." % (self.table.name, detail)
+                )
             elif constraint_name_string(const.name):
                 self.named_constraints[const.name] = const
             else:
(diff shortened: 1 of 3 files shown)


sqla-tester:

This pull request references issue #1846, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #1846, reopen this pull request and it will stay open. If #1846 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

PR #1862, opened 2026-09-01
docs: document SQLite batch migrations with views

### LINKED ISSUE #1207: Running `alter table` revisions w/batch operations on SQLite raises an error if table referenced in a view
**Describe the bug**
*This may be not so much an Alembic bug as a limitation of SQLite; nevertheless, the following use case results in an error that is not mentioned in the documentation; besides, there seems to be a relatively simple fix for this.*

Consider the following scenario:
1. create table foo
2. create view myfoo
3. add a revision that adds a column from foo
4. run upgrade, then downgrade. The downgrade will raise an error.

The specific schema of foo, as well as the column we are adding/dropping are irrelevant. But here's an example:

```py
########################## REVISION 1 ########################## 
def upgrade() -> None:
    op.create_table("foo", sa.Column("id", sa.Integer, primary_key=True))   

def downgrade() -> None:
    op.drop_table("foo")

########################## REVISION 2 ##########################
def upgrade() -> None:                                                                                                                                                                        
    op.execute("CREATE VIEW my_foo AS SELECT id FROM foo")                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                              
def downgrade() -> None:                                                                                                                                                                      
    op.execute("DROP VIEW my_foo")

########################## REVISION 3 ##########################
def upgrade() -> None:
    with op.batch_alter_table("foo") as batch_op:
        batch_op.add_column(sa.Column("stuff", sa.String))

def downgrade() -> None:
    with op.batch_alter_table("foo") as batch_op:
         batch_op.drop_column("stuff")

# (I can provide the full revision files if needed)
```
Running migrations results in the following:
```sh
$ alembic upgrade heads
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade  -> 0a35dd0c400b, create table foo
INFO  [alembic.runtime.migration] Running upgrade 0a35dd0c400b -> 70ec189d8dea, create view myfoo
INFO  [alembic.runtime.migration] Running upgrade 70ec189d8dea -> c907a34d4a26, add column to table foo

$ alembic downgrade -1
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
INFO  [alembic.runtime.migration] Running downgrade c907a34d4a26 -> 70ec189d8dea, add column to table foo
  File "/home/jdavcs/0dev/drafts/python/sqlalchemy/views/v2/.venv/lib64/python3.9/site-packages/sqlalchemy/engine/base.py", line 1964, in _exec_single_context
    self.dialect.do_execute(
  File "/home/jdavcs/0dev/drafts/python/sqlalchemy/views/v2/.venv/lib64/python3.9/site-packages/sqlalchemy/engine/default.py", line 748, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: error in view my_foo: no such table: main.foo
... [output truncated]
[SQL: ALTER TABLE _alembic_tmp_foo RENAME TO foo]
```

I believe this is happening primarily due to how SQLite handles the [`ALTER TABLE`](https://sqlite.org/lang_altertable.html) statement. The error can be easily reproduced by reducing the logic of the above revisions to the following SQL statements (which is a simplified version of what alembic does):

```sql
CREATE TABLE foo (id INTEGER, data INTEGER);
CREATE VIEW fooview AS SELECT id FROM foo;
/* here we drop the `data` column from foo by altering the table via copy/drop/rename */
CREATE TABLE tmp_foo (id INTEGER);
/* skipped step as irrelevant: here we'd copy the rows from foo into tmp_foo */
DROP TABLE foo;
ALTER TABLE tmp_foo RENAME TO foo;  /* Error: error in view fooview: no such table: main.foo */
```

**Expected behavior**
No error, OR a note in the documentation on [batch migrations](https://alembic.sqlalchemy.org/en/latest/batch.html#running-batch-migrations-for-sqlite-and-other-databases) explaining that if a view references the table being altered, and SQLite's version is < 3.35, an error will happen.

**To Reproduce**
See description. SQLite version: 3.34 (must be < 3.35, as 3.35 introduced the `ALTER TABLE DROP COLUMN` statement.)

**Possible Solution**
As per SQLite's documentation, enabling the [`legacy_alter_table`](https://sqlite.org/pragma.html#pragma_legacy_alter_table) setting should solve this. Indeed, adding the `PRAGMA legacy_alter_table=1` statement to the example above appears to solve the problem. The following code runs without errors, making the correct changes to the database:

```sql
CREATE TABLE foo (id INTEGER, data INTEGER);
CREATE VIEW fooview AS SELECT id FROM foo;
CREATE TABLE tmp_foo (id INTEGER);
DROP TABLE foo;
PRAGMA legacy_alter_table=1;
ALTER TABLE tmp_foo RENAME TO foo;
PRAGMA legacy_alter_table=0
```

**Versions.**
 - OS: Fedora 34
 - Python: 3.9.12
 - Alembic: 1.10.2
 - SQLAlchemy: 2.0.7 (irrelevant)
 - Database: SQLite: 3.34.1

Ryota-Di:
**This project accepts pull requests only for issues that a maintainer has
marked with the `open for pull requests` label.**

A pull request that doesn't reference such an issue **is closed
automatically**. That isn't a judgment on your change: it's how we settle on
an approach before anyone spends time writing code, and how we keep two
people from doing the same work at once.

The issue you reference has to be:

* in this repository,
* open,
* labeled `open for pull requests`, and
* not already labeled `code review in progress`, which means someone is
  working on it already.

If there's no such issue yet, open one describing the problem or the feature,
including a complete, runnable example, and wait for a maintainer to add the
label. Please don't open a pull request first and ask for the label
afterwards.

### Fixes

Fixes: #1207

### Description

Documents a SQLite limitation when using batch migrations with views that
reference the table being recreated.

During the "move and copy" workflow, the original table is temporarily
dropped before the replacement table is renamed. SQLite may reject the
rename when an existing view still refers to the temporarily missing table.

The documentation also notes that temporarily enabling
`PRAGMA legacy_alter_table` while the replacement table is renamed can be
used as a workaround.

Verification:

- `git diff --check`
- Sphinx documentation build with warnings treated as errors
- Reproduced the failure locally with SQLite 3.53.1
- Verified that temporarily enabling `PRAGMA legacy_alter_table` allows the
  batch migration to complete and leaves the referencing view usable

### DIFF
--- a/docs/build/batch.rst
+++ b/docs/build/batch.rst
@@ -272,6 +272,19 @@ by a batch migration.  Therefore it may be necessary to manipulate the
 ``PRAGMA FOREIGN KEYS`` setting if a migration seeks to rename a table vs.
 batch migrate it.
 
+.. _batch_referencing_views:
+
+Dealing with Referencing Views
+------------------------------
+
+On SQLite, batch migrations that use the "move and copy" workflow may fail
+when a view references the table being altered.  The workflow temporarily
+drops the original table before renaming the replacement table, and SQLite
+may reject the rename while the view refers to the missing table.
+
+As a workaround, SQLite's ``PRAGMA legacy_alter_table`` setting may be
+temporarily enabled while the replacement table is renamed.
+
 .. _batch_offline_mode:
 
 Working in Offline Mode


sqla-tester:
Thanks! Issue #1207 is now marked **code review in progress** and no longer **open for pull requests**, so this pull request holds the review for it and another one won't land on top of your work. If this pull request is abandoned, a maintainer can put **open for pull requests** back on #1207 to reopen it to others.

PR #1863, opened 2026-09-02
fix(autogenerate): exclude type-bound check constraints from reflected comparison

### LINKED ISSUE #1859: Autogenerate reports false remove_constraint for Enum/Boolean(create_constraint=True) columns
**Describe the bug**
The check-constraint autogenerate detection added in 1.19.0 (#508) reports a false `remove_constraint` for CHECK constraints generated by `Enum`/`Boolean(create_constraint=True)`, even when the table was created directly from the same `MetaData`, i.e., there is no drift at all.

This looks related to #1842 (fixed in 1.19.1), but is distinct: #1842 was about an explicit `CheckConstraint` passed as a *column argument*. This is about the constraint a `SchemaType` (`Enum`/`Boolean`) generates itself, which is still misdetected on 1.19.1.

**Expected behavior**
`diff == []`. The table was created directly from this `MetaData`, so there's nothing to detect.

**To Reproduce**
```python
from sqlalchemy import Column, Enum, Integer, MetaData, Table, create_engine
from alembic.autogenerate import compare_metadata
from alembic.migration import MigrationContext

metadata = MetaData()

Table(
    "widget",
    metadata,
    Column(
        "status",
        Enum(
            "active",
            "inactive",
            name="widget_status",
            native_enum=False,
            create_constraint=True,
        ),
        nullable=False,
    ),
)

engine = create_engine("sqlite://")
metadata.create_all(engine)  # table created from *this exact* metadata

with engine.connect() as connection:
    context = MigrationContext.configure(connection)
    diff = compare_metadata(context, metadata)

print("diff:", diff)
assert diff == []
```

**Error**

```
diff: [('remove_constraint', CheckConstraint(<sqlalchemy.sql.elements.TextClause object at 0x...>, name='widget_status', table=Table('widget', MetaData(), Column('x', Integer(), table=<widget>), schema=None)))]
AssertionError
```

Confirmed as a 1.19 regression by bisecting on this same script:

| alembic | result |
|---|---|
| 1.18.5 | `diff: []` (pass) |
| 1.19.0 | false `remove_constraint` |
| 1.19.1 | false `remove_constraint` (still present after the #1842 fix) |

Also confirmed dialect-independent: the same false positive reproduces
against a real PostgreSQL connection, not just SQLite.

**Versions.**
- alembic: 1.19.1
- SQLAlchemy: 2.0.52
- Python: 3.13.11
- Database: reproduces on both SQLite and PostgreSQL 18

**Additional context**
It looks like the root cause is in `alembic.util.sqla_compat.all_table_check_constraints()`, i.e., the helper the #1842 fix introduced. It still excludes type-bound constraints:

```python
def all_table_check_constraints(table: Table) -> set[CheckConstraint]:
    """Returns all check constraint that are not type-bound,
    including those on columns."""
    candidates = chain(
        table.constraints, *(c.constraints for c in table.columns)
    )
    return {
        ck
        for ck in candidates
        if isinstance(ck, CheckConstraint) and not _is_type_bound(ck)
    }
```

`_is_type_bound` returns `True` for constraints SQLAlchemy generates from a
column's type (`Enum`/`Boolean` with `create_constraint=True`), so they're
never added to the metadata-side comparison set -- while the new
`checkconstraint_byname` autogenerate plugin *does* reflect them from the
live database. Every reflected type-bound constraint therefore comes back
as "should be removed," with no way for the comparator to reconcile it
against metadata.

Related: #508 (introduced the feature), #1842 (adjacent, already-fixed column-argument case).

**Have a nice day!**

mturac:
## Summary

- `Enum`/`Boolean` columns with `create_constraint=True` generate type-bound check constraints
- The metadata side correctly excludes these via `all_table_check_constraints()`, but the reflection side still returns them from the database
- This causes false `remove_constraint` operations even when the table was created from the exact same metadata

## Changes

- Added `type_bound_check_constraint_names()` in `sqla_compat.py` to collect names of type-bound constraints from metadata
- Filter reflected constraints by these names in `_compare_check_constraints()` before comparison
- Added tests for both Boolean and Enum type-bound constraint scenarios

## Test plan

- [x] `test_existing_type_bound_boolean_not_detected` — verifies no false diff for Boolean(create_constraint=True)
- [x] `test_existing_type_bound_enum_not_detected` — verifies no false diff for Enum(create_constraint=True, native_enum=False)
- [x] Existing check constraint tests still pass

Closes #1859


sqla-tester:

This pull request references issue #1859, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #1859, reopen this pull request and it will stay open. If #1859 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

zzzeek:
cc @CaselIT

CaselIT:
it doesn't really solves anything, since the issue with unnamed constraint in the metadata that came back named from the db is not solved ad all

PR #1869, opened 2026-09-15
fix(postgresql): identity column not added when existing_server_default omitted

### LINKED ISSUE #1504: Autogeneration causes invalid SQL with some PostGRES serial columns
**Describe the bug**
Auto-generated PostgreSQL code in 1.13.2 contains invalid code causing migrations to fail. This appears to be around Serial columns and appears to be related to the fix implemented in: https://github.com/sqlalchemy/alembic/issues/1479. I get the log around detecting a serial and omitting it. We have a ton of tables like this, and not all of them are having this behavior. The table definitions haven't changed, and everything works in 1.13.1.

The id columns definitely have a server default set, but all of them do, not just the ones triggering this behavior. I am having trouble figuring out what is different between them. The python definition is the same, and the database definitions appear to be almost identical.

I tried replicating it in a fresh database to generate a minimum code sample, but couldn't replicate the behavior

**Expected behavior**
Expect no change from previous behavior.

**To Reproduce**
Table Definition:
```py
class GEO(Base):
    __tablename__ = "GEO"
    id: Mapped[int] = mapped_column(Identity(on_null=True), primary_key=True)
    name: Mapped[str] = mapped_column(String(4), unique=True)
```

```py
from typing import Sequence, Union

from alembic import op
import sqlalchemy as sa


# revision identifiers, used by Alembic.
revision: str = 'b6001b86cfc9'
down_revision: Union[str, None] = 'd4e45e048b78'
branch_labels: Union[str, Sequence[str], None] = None
depends_on: Union[str, Sequence[str], None] = None


def upgrade() -> None:
    op.alter_column('GEO', 'id',
               existing_type=sa.INTEGER(),
               server_default=sa.Identity(always=False, on_null=True),
               existing_nullable=False,
               autoincrement=True)



def downgrade() -> None:
    op.alter_column('GEO', 'id',
               existing_type=sa.INTEGER(),
               server_default=None,
               existing_nullable=False,
               autoincrement=True)

```

**Error**

Generated SQL it tries to run during migration is: `ALTER TABLE "GEO" ALTER COLUMN id; `
Which doesn't do anything and isn't valid and throws an error.

**Versions.**
 - OS: Docker in docker container
 - Python: 3.11
 - Alembic: 1.13.2
 - SQLAlchemy: 2.0.31
 - Database: RDS PostGRES 14.10
 - DBAPI: psycopg2

RitiGrover:
## Description
`Operations.alter_column` defaults `existing_server_default` to `False` (its internal "not specified" sentinel) when the caller omits it. The PostgreSQL identity-column DDL compiler (`visit_identity_column` in `alembic/ddl/postgresql.py`) only checked `is None` to decide whether to render `ADD GENERATED ... AS IDENTITY`, so `False` fell through into the "diff against an existing identity" branch, computed an empty diff against the sentinel, and rendered an invalid, clause-less `ALTER TABLE ... ALTER COLUMN id;`.

## Motivation and Context
Fixes #1504. Any migration that sets `server_default=Identity(...)` on `alter_column` without also passing `existing_server_default` (the common case, since there usually isn't a prior identity to describe) fails at execution time with a Postgres syntax error.

## How Has This Been Tested?
- Reproduced against a real PostgreSQL 16 instance: confirmed the exact invalid SQL from the issue, confirmed the fix resolves it
- Verified the two related paths still work correctly against the live DB: altering an existing identity's options, and dropping identity
- Added a regression test (`test_add_identity_to_column_no_existing_server_default`) mirroring the existing `test_add_identity_to_column` test; confirmed it fails on the old code and passes with the fix
- Ran the full `test_postgresql.py` identity suite (13/13 pass) plus the rest of the offline and live-DB PostgreSQL suite - no regressions
- `flake8`, `black --check`, `mypy` all clean on the changed files


sqla-tester:

This pull request references issue #1504, which hasn't been marked **open for pull requests**, so the change isn't authorized yet and I'm closing it automatically. Once a maintainer adds the label to #1504, reopen this pull request and it will stay open. If #1504 needs more detail before that can happen, a complete runnable example is usually the missing piece, and adding one to the issue is the fastest way to get there.

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **zzzeek** to try to get revision e3d5339d6ee119833863d381c31f8ba59e0cd71d of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
New Gerrit review created for change e3d5339d6ee119833863d381c31f8ba59e0cd71d: https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6982

zzzeek:
OK, it failed.  the "on_null" thing does not seem to be platform agnostic.   if you're using an LLM, make sure it runs the tests before submitting

tests/test_postgresql.py
```diff
@@ -449,6 +449,24 @@ def test_remove_identity_from_column(self):
             "ALTER TABLE t1 ALTER COLUMN some_column DROP IDENTITY"
         )
 
+    def test_add_identity_to_column_no_existing_server_default(self):
```
sqla-tester: **Federico Caselli** (CaselIT) wrote:

how about we add a matrix to test_add_identity_to_column to control existing_server_default?

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6982

sqla-tester (commented):

change seems fine, but we could likely reuse an existing test for this

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6982

RitiGrover:
Fixed - thanks a lot for your review

sqla-tester (commented):
OK, this is **sqla-tester** setting up my work on behalf of **CaselIT** to try to get revision 6e594213ba55b9ae9c5a08eca3dfe62474f4f284 of this pull request into gerrit so we can run tests and reviews and stuff

sqla-tester:
Patchset 6e594213ba55b9ae9c5a08eca3dfe62474f4f284 added to existing Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6982

sqla-tester (commented):

code review left on gerrit

View this in Gerrit at https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6982

* tests/test_postgresql.py (line 452): Done

zzzeek:
I made some final adjustments.   thanks!

sqla-tester:
Gerrit review https://gerrit.sqlalchemy.org/c/sqlalchemy/alembic/+/6982 has been **merged**. Congratulations! :)

